PR #9, opened 2018-09-03
fix box zoom in julia 1

piever:
The new `ntuple` syntax is `ntuple(Val(n), f)`


SimonDanisch:
Thanks, much appreciated! I actually just also pushed a fix, though: 6d9ec40dbb3d2bd738f367822e28bceaed7b6ab8 :D

PR #12, opened 2018-09-03
fix recipe docs

piever:
Unless I'm misunderstanding how this works, this macro requires the type rather than the function here.


SimonDanisch:
Thank you!

PR #18, opened 2018-09-03
Axis texts in black, no additional axes through origin in 2D

mschauer:
cc @mkborregaard  

Make axis texts black. 
Currently, axes within the picture through the origin aren't well supported in 2D plots, so setting `axis_position = nothing` makes most sense at the moment. 
Adresses https://github.com/JuliaPlots/Makie.jl/issues/129

### DIFF
--- a/src/basic_recipes/axis.jl
+++ b/src/basic_recipes/axis.jl
@@ -51,7 +51,7 @@ using .Formatters
             linewidth = 1.0,
             linecolor = :black,
             linestyle = nothing,
-            axis_position = :origin,
+            axis_position = nothing,
             axis_arrow = false,
             arrow_size = 2.5,
             frames = ((false, false), (false, false)),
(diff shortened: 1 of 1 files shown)


src/basic_recipes/axis.jl
```diff
@@ -125,7 +123,7 @@ end
         frame = Theme(
             linecolor = (grid_color, grid_color, grid_color),
             linewidth = (grid_thickness, grid_thickness, grid_thickness),
-            axiscolor = (darktext, darktext, darktext),
```
mschauer: This has no 2d equivalent?

SimonDanisch: ah, I'm actually also not using it in the code :D this needs a clean up at some point ;)

mschauer: You are using it, I am starting to understand what's going on. Cleanup and unifying a bit 2d and 3d interface would be good, I'll keep my eyes on it and contribute.

SimonDanisch:
Uh, @mschauer is this working fine? I'd just merged it if it is ;)

mschauer:
Yeah, it basically preserves the disorder...

SimonDanisch:
Nice! :D

PR #19, opened 2018-09-04
define convert arguments for functions

piever:
This implements Plots recipes for functions, using the `adapted_grid` algorithm to detect the optimal set of points on the `x` axis.

Now for examples `lines(sin, -pi, pi)` gives:

test


SimonDanisch:
Beautiful, thanks!

PR #21, opened 2018-09-05
remove deprecated ind2sub

piever:
Related question: don't we need a `grid!` method as well? Otherwise how can one use the `AbstractPlotting.plot!` method to define a recipe that needs some layouting?


SimonDanisch:
We'll need quite a bit for layouting... I was thinking to wait for the Observable changes, since that might make it easier to get the resizing layouts to work...

PR #29, opened 2018-10-09
Fix Orthographic projection

ffreyer:
In relation to https://github.com/JuliaPlots/Makie.jl/issues/156

The issue seems to be the call to `GLAbstraction.to_worldspace`. Rotation, translation, zoom and resizing all work fine without it (and an orthographic camera). Though the initial zoom level is very zoomed in.


SimonDanisch:
Huh, that's weird that this deletion _fixes_ it ... Sorry for the slow reply, but that seems odd - I'll need to investigate!

SimonDanisch:
Ok, this is weird, I'll need to figure out how to do it properly at some point.
But for now this is a good enough fix, thank you!

PR #31, opened 2018-10-30
implement `frames` axis argument

mkborregaard:
Allows
```julia
using AbstractPlotting, Makie
newtheme = Theme(
    axis = Theme(
        frame = Theme(
            linecolor = :red,
            frames = ((true, false), (true, false))
        )
    )
)


AbstractPlotting.set_theme!(newtheme)
scene = scatter(rand(100), rand(100))
```
skaermbillede 2018-10-30 kl 09 21 48

I'm not so sure the ordering of `true`s and `false`s is intuitive but not sure what would be either.

### DIFF
--- a/src/basic_recipes/axis.jl
+++ b/src/basic_recipes/axis.jl
@@ -245,7 +245,7 @@ end
 function draw_frame(
         linebuffer, limits::NTuple{N, Any},
         linewidth, linecolor, linestyle,
-        axis_position, axis_arrow, arrow_size
+        axis_position, axis_arrow, arrow_size, frames
     ) where N
 
     mini = minimum.(limits)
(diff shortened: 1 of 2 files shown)


mkborregaard:
Note that the default `frames` is currently `((false, false), (false, false))` so this PR will remove the frames from default plots. Should that default be changed?

mkborregaard:
I've added the possibility of doing `frames = :semi` (`:box` and `:none` (the current default)) instead of having to pass the double tuples.

SimonDanisch:
Awesome :) is this ready to merge?

mschauer:
I think the `semi` style should actually not be implemented as part of the frame but correspond to `frames = :none` and a corresponding value for `axis_position` like `:standard`, `:leftbottom` or something similar.

mkborregaard:
Right now `frames` double as axis lines, no? I've simply used the frame bits as axis lines.

But in fact I do agree it would make sense to have an axis that is a line associated with the tick marks as something distinct from the frame (sometimes with an arrowhead), but USUALLY completely overlapping the frame. See examples of dichotomies between frames and axis lines in Mathematica here:

billede

mkborregaard:
We could still keep the semi axis style, and keep this PR as is, though. That allows dichotomy between lower left axis lines and no box:

billede

and lower left axis lines and a box:
billede

mschauer:
Exactly, this is even partly supported already. There is `axis_position = :origin` which draws an axis insight the picture if the origin is insight the drawing area.

mkborregaard:
I just didn't see where the axis line is drawn?

mkborregaard:
Anyway I think the conclusion still is that this is ready to merge IMHO.

SimonDanisch:
great :) thanks for fixing this nicely!

SimonDanisch:
I think we can still bike-shed those details in other PRs :)

mkborregaard:
Yes I'm happy to take a look at the axis suggested by @mschauer next (though "next" is a wide term).

PR #35, opened 2018-10-30
recipes improvements

SimonDanisch:


SimonDanisch:
This is how a qqpair recipe could look like:
```julia
function recipe!(scene::Combined, P::PlotType, qqpair::QQPair; qqline = :identity, kw_args...)
    scene.axis.labels = ("qx", "qy")
    plot!(scene, P, qqpair.qx, qqpair.qy)
    if qqline == :identity
        lines!(scene, 1:10, 1:10)
    end
end
```
Btw, if we want to access e.g. axis limits etc, while the scene isn't really there yet in this stage of the recipe, we could always return a Node - basically making it lazy and then updating the node once the whole scene is composed! So e.g. the following could work:
```julia
lines!(scene, lift(x-> range(minimum(x), stop = maximum(y), length = length(qqline)), limits(scene))
```

piever:
I just love it that `qqpair` became the universal recipe example :)

I also really like this approach to recipes!

For me to understand: if the user passes explicit axes labels, say:

```julia
plot(my_qqpair, axis = (; names = (; axisnames = ("a", "b"))))
```

would that overwrite the qqpair default? I think it should, Plots gives you the option in a recipe (`-->` vs `:=` I think) to either set a "soft default" (overridable) or a "hard default" (ignore user input), but I think soft default is more generally useful.

SimonDanisch:
`scene.axis.labels = ("qx", "qy")` will only set the names inside the plot object, which the composing recipe is free to use to _actually_ set the axis labels ;) So, the user can still overwrite the names, since `scene.axis.labels = ("qx", "qy")` will get treated as just a "recommendation" !

PR #40, opened 2018-11-08
WIP: allow to use arguments in default theme

piever:
The `default_theme` function now also gets passed the converted arguments, which allows to customize the theme as a function of the arguments the user is passing , for example:

```julia
density(Surface, rand(100,2))
```

could be instructed to use `Surface` theme.

### DIFF
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -5,9 +5,9 @@ const Theme = Attributes
 
 Theme(x::AbstractPlot) = x.attributes
 
-default_theme(scene, T::Type, args...) = Attributes()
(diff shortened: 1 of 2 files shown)


piever:
I'm not particularly convinced by the implementation of the PR, but I hope it illustrates the idea (as a side effect, it causes the axis frame to disappear, not sure why).

piever:
For it to be non-breaking the function passed to the macro discards the keyword arguments by default, but maybe there's a smarter way to avoid having to write the function twice.

piever:
Closing as after the StatsMakie redesign as well as #46 this doesn't seem relevant anymore.

PR #42, opened 2018-11-09
Lifted convert_arguments + attribute propagation

SimonDanisch:


src/interfaces.jl
```diff
@@ -398,12 +487,45 @@ eval(default_plot_signatures(:plot, :plot!, :Any))
...
-    PlotType = plottype(to_value.(args)...)
-    plot!(scene, PlotType, attributes, args...)
+function plot!(scene::SceneLike, P::PlotFunc, attributes::Attributes, args...; kw_attributes...)
+    attributes = merge!(Attributes(kw_attributes), attributes)
+    argvalues = to_value.(args)
+    PreType = plottype(P, argvalues...)
+    # plottype will lose the argument types, so we just extract the plot func
+    # type and recreate the type with the argument type
+    PreType = Combined{plotfunc(PreType), typeof(argvalues)}
+    keys = used_attributes(PreType, argvalues...)
```
piever: maybe `keys = intersect(used_attributes(PreType, argvalues...), keys(attributes))` ? Otherwise we throw unless the user provides all keywords

SimonDanisch: keys should be immutable ;) so we'd need to filter them out a step later!

SimonDanisch: ah, but the number of passed attributes  wont change either^^ so, guess that should work :)

examples/function.jl
```diff
@@ -0,0 +1,27 @@
...
+to_tuple(t) = (t,)
+
+
+function convert_arguments(P::PlotFunc, f::Function, args...; kwargs...)
+    tmp = f(args...; Iterators.filter(t -> last(t) != automatic, kwargs)...) |> to_tuple
+    convert_arguments(P, tmp...)
+end
+
+# remove convert_arguments(P, f, x) = (x, f.(x))
+function convert_arguments(P::PlotFunc, d::KernelDensity.UnivariateKDE)
```
piever: Why `PlotFunc` here and `Type{<:Combined{T}}` below?

examples/function.jl
```diff
@@ -0,0 +1,27 @@
...
+    ptype = plottype(P, Lines) # choose the more concrete one
+    ptype => convert_arguments(ptype, d.x, d.density)
+end
+
+function convert_arguments(P::Type{<: Combined{T}}, d::KernelDensity.BivariateKDE) where T
+    ptype = plottype(P, Heatmap)
```
piever: I'm not sure I understand  `plottype`. I though it would somehow figure out how to combine the plot passed as a function and the automatic plot, but it doesn't always work, for example `plottype(Lines, Lines) != Lines`

examples/function.jl
```diff
@@ -0,0 +1,27 @@
...
+    ptype => convert_arguments(ptype, d.x, d.y, d.density)
```
piever: Is `convert_arguments` always expected to return a pair at the end? Or is there a `to_pair` function that would turn it into a pair with the first element the "undefined plot" (I guess `Combined{Any, <:Any}`)?

piever:
I love this PR! Have been using it for implementing `Group` and it makes things much easier. I've added comments for things I don't understand.

One thing that you may have noticed already: this goes 90% of the way in terms of "lifted recipes", because one can simply use a `convert_argument` method to return an array of series and add a recipe to plot array of series: it's how I've implemented `Group` so far, see https://github.com/JuliaPlots/StatsMakie.jl/pull/7/files#diff-dcc8ed379e5563228673316edce385c9R49 for example (`convert_arguments` returns a `PlottableTable` where every row is a "series"). Of course the specific recipe to plot the array of series needs some extra work (I give up on interactivity on mine) and we need to think carefully on how it handles attributes, because it is the place where we can introduce smart syntax like in Plots when you pass an attribute across series with row vectors.

examples/function.jl
SimonDanisch: yeah maybe I shouldn't overload this functionality for `plottype` - it basically just figures out the most concrete plot type... so if you have `plottype(Combined{Any}, Lines) == Lines`, but when it's `plottype(Scatter, Lines) == Scatter)`

examples/function.jl
SimonDanisch: No, the pair is optional, but you want to define a default plot type for BivariateKDE, so you need to return a pair with the plot type;)

SimonDanisch:
Yeah that's what I meant when I said:
```
btw, the plan for me right now is to just get this [lifted recipes] working nicely, and later we can extend it to also allow multiple series + layouts, simply by adding a scene recipe for the output type of convert arguments
```
:)

examples/function.jl
piever: Then maybe it'd be nice to have a `to_pair(P, args)` where `P` is the plottype that comes explicitly from the user that does the right thing (creates a pair plottype => args), taking into account whether `args` itself is a pair or not. And maybe even a `to_pair(args) = to_pair(Combined{Any, <:Any}, args)`

piever:
I see, it took me a little bit longer to get there but now I understand the design :) 

I'll keep my temporary implementation of this in the StatsMakie PR then (under the creative name `PlottableTable`) to see exactly what features should this recipe have to work well with the grouping framework.

examples/function.jl
SimonDanisch: Ok, maybe the name `to_pair` throws me off... I guess you're talking about the possibility of recursively applying convert_arguments? If so, can you come up with an example, were you'd want to use convert_arguments recursively and not directly return the result?

piever: It was actually in an attempt to implement a "basic recipe" for the `SeriesVector`, where each entry was basically a collection of attributes and the return type of `convert_arguments`. The logic becomes much clearer if I know that each of those is a pair (pretty much like we do `to_node` to support both `Node` and anything else by converting everything to a `Node`). OTOH the name `to_pair` is not perfect, but I can't think of a much better one...

Here for example I find `ptype => convert_arguments(ptype, d.x, d.density)` a bit puzzling. What if the second call also returns a `Pair`? Do you end up with `ptype => P => args`?

>  If so, can you come up with an example, were you'd want to use convert_arguments recursively and not directly return the result?

With `Grouping` it's quite common as I'd split the data in separate groups, call `convert_arguments` in all the groups and then I'd like to figure out what the overall plottype is and create a structure with only the actual plot arguments. This could maybe be handled by creating a recipe for:

```
plot!(::Combined{T, Tuple{Pair{<:PlotFunc, S}}})
```

so that in the end it doesn't matter if I pass the pair or the arguments.

src/interfaces.jl
```diff
@@ -398,12 +487,45 @@ eval(default_plot_signatures(:plot, :plot!, :Any))
...
+        lift((args...)-> Pair.(keys, args), getindex.(attributes, keys)...) # make them one tuple to easier pass through
+    end
+    # call convert_arguments for a first time to get things started
+    converted = convert_arguments(PreType, argvalues...; kw_signal[]...)
+    # convert_arguments can return different things depending on the recipe type
+    # apply_conversion deals with that!
+    FinalType, argsconverted = apply_convert!(PreType, attributes, converted)
+    converted_node = Node(argsconverted)
+    input_nodes =  to_node.(args)
+    onany(kw_signal, lift(tuple, input_nodes...)) do kwargs, args
```
piever: Somehow I feel here it'd be good to add some "Trait" to these functions (meaning, the "listeners"). I think that there will be several occasions where the user doesn't want to call `convert_arguments` again when just one argument changes. Say for example that I'm calling it on `x,y` that have to be of equal length. If I make `x` longer I'd like to wait that I've also made `y` shorter. So I would add a "is_convert_argument" trait to these updater so that the user can do:

```julia
setindex!(x, x_new, notify = !is_convert_argument)
setindex!(y, y_new)
```

PR #43, opened 2018-11-10
function conversion cleanup

piever:
Now the allowed signatures are:

```julia
# x is AbstractInterval, y is function
julia> lines(-pi..pi, sin)

# x is AbstractVector, y is function
julia> lines(-pi:0.1:pi, sin)
````

### DIFF
--- a/src/basic_recipes/basic_recipes.jl
+++ b/src/basic_recipes/basic_recipes.jl
@@ -414,8 +414,8 @@ end
 
 convert_arguments(P::Type{<:AbstractPlot}, r::AbstractVector, f::Function) = convert_arguments(P, r, f.(r))
 
-function convert_arguments(P::Type{<:AbstractPlot}, i::Interval, f::Function)
-    convert_arguments(P, PlotUtils.adapted_grid(f, extrema(i)), f)
+function convert_arguments(P::Type{<:AbstractPlot}, i::AbstractInterval, f::Function)
+    convert_arguments(P, PlotUtils.adapted_grid(f, endpoints(i)), f)
 end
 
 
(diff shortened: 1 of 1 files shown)


mkborregaard:
What about just `lines(sin)`?

SimonDanisch:
I was against that, as long as we dont employ an algorithm that figures out the lims of `f`... a hardcoded range seems pretty arbitrary and I'm not entirely sure what use case it's supposed to cover ;)

piever:
I would tend to agree and keep the one argument methods only for the case where there is a good heuristic to determine the `x` axis (for example `plot(Normal())` where one ca use the quantiles to get the extrema).

mkborregaard:
Just convenience I guess. It's nice to just throw your function at `plot` and see it work.
It doesn't have to be a hardcoded range. In Plots we use the current window limits if you're adding to an existing plot, otherwise it tries to work out the domain over which the function is defined and acts over that : https://github.com/JuliaPlots/Plots.jl/blob/45b5b33929c7454d15e4477296c09f3527d8018c/src/series.jl#L382-L414

So that e.g. `plot(x->sqrt(-x))` gets plotted on `(-5, 0)`.

piever:
I think we should wait until we have a clear idea how to do that and here I wanted to limit the proposal to cases where the behavior is very clear.

In the `plot(f)` case for example I'm not even sure we should assume that the function takes one argument: if it only has a method with two arguments I would actually try and make a `surface` by default: it'd be interesting to start a separate PR and discuss heuristics.

mkborregaard:
👍 makes sense

mkborregaard:
I know it's a bit late to comment on a merged branch, but why `lines(-pi..pi, sin)` and not `lines(sin, -pi..pi)`, given that it's actually the function we are plotting?

piever:
The reasoning is that the method `plot(f, args...)` means "call `f` on `args` and plot the result" and is used in the Grammar of Graphics, for example `plot(kde, rand(100))`. Instead `lines(-pi..pi, sin)` is in line with `heatmap(x, y, f)`, where `f` represents the `z` axis so stays in the "z" position.

mkborregaard:
OK - but why would f necessarily be z? In a heatmap, f controls z, yes, but that's not the case for `lines`. I guess you could argue that it controls `y` that should then still come after `x`...

PR #45, opened 2018-11-15
generalize palette

piever:
This adds support for a generic palette type that gets automatically converted to one of its values on `convert_attributes` (either the first value or a cycling one, depending whether it's a cycling palette or not). I'll probably also add some default palettes for the attributes we want to use in Grammar of Graphics (marker, linestyle etc).

### DIFF
--- a/src/theming.jl
+++ b/src/theming.jl
@@ -2,7 +2,7 @@
 const minimal_default = Attributes(
     font = "Dejavu Sans",
     backgroundcolor = RGBAf0(1,1,1,1),
-    color = :black,
+    color = Palette(ColorBrewer.palette("Dark2", 8), cycle = true),
     colormap = :viridis,
     resolution = reasonable_resolution(),
     visible = true,
--- a/src/AbstractPlotting.jl
+++ b/src/AbstractPlotting.jl
@@ -25,6 +25,7 @@ include("interaction/nodes.jl")
 
 # Basic scene/plot/recipe interfaces + types
 include("scenes.jl")
+include("palette.jl")
 include("theming.jl")
 include("recipes.jl")
 include("interfaces.jl")
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -363,18 +363,6 @@ function convert_arguments(
     (m,)
 end
 
-struct Palette{N}
-   colors::SArray{Tuple{N},RGBA{Float32},1,N}
-   i::Ref{UInt8}
-   Palette(colors) = new{length(colors)}(SVector{length(colors)}(to_color.(colors)), zero(UInt8))
-end
-Palette(name::Union{String, Symbol}, n = 8) = Palette(to_colormap(name, n))
-
-function convert_attribute(p::Palette{N}, ::key"color") where {N}
-    p.i[] = p.i[] == N ? one(UInt8) : p.i[] + one(UInt8)
-    p.colors[p.i[]]
-end
-
 convert_attribute(c::Colorant, ::key"color") = convert(RGBA{Float32}, c)
 convert_attribute(c::Symbol, k::key"color") = convert_attribute(string(c), k)
 function convert_attribute(c::String, ::key"color")
--- a/src/palette.jl
+++ b/src/palette.jl
@@ -1,0 +1,21 @@
+struct Palette{S, T<:AbstractVector{S}} <: AbstractVector{S}
+   values::T
+   i::Ref{UInt8}
+   cycle::Bool
+   Palette(values::T; cycle = false) where {T<:AbstractVector} =
+       new{eltype(T), T}(values, Ref{UInt8}(1), cycle)
+end
+
+Palette(name::Union{String, Symbol}, n = 8; kwargs...) = Palette(to_colormap(name, n); kwargs...)
+
+for s in [:(Key), :(Key{:color})]
+    @eval function convert_attribute(p::Palette, key::($s))
+        attr = convert_attribute(p.values[p.i[]], key)
+        p.cycle && (p.i[] = p.i[] == length(p.values) ? one(UInt8) : p.i[] + one(UInt8))
+        attr
+    end
+end
+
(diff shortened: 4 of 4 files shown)


src/theming.jl
```diff
@@ -2,7 +2,7 @@
 const minimal_default = Attributes(
     font = "Dejavu Sans",
     backgroundcolor = RGBAf0(1,1,1,1),
-    color = :black,
+    color = Palette(ColorBrewer.palette("Dark2", 8), cycle = true),
```
mkborregaard: Is this necessary? Why not just `Palette(:Dark2)`?

piever: Because this is executed at precompilation time, so the `Palette(:Dark2)` method is not yet available. From the user side the two calls are equivalent.

mkborregaard:
I was just making a PR to add Dark2 as the default color, all you need in addition is to export `Palette` so I've dropped my PR. But note that there is currently a problem with the Palettes - which is that a new `scene` doesn't reset the `Theme` components. This leads to this:
```julia
    scene = scatter(randn(20), randn(20), markersize = 0.2)
    scatter!(scene, randn(20), randn(20), markersize = 0.2)
```
skaermbillede 2018-11-15 kl 18 12 59

and the second time
```julia
    using Makie
    scene = scatter(randn(20), randn(20), markersize = 0.2)
    scatter!(scene, randn(20), randn(20), markersize = 0.2)
```
gives
skaermbillede 2018-11-15 kl 18 13 19

piever:
> But note that there is currently a problem with the Palettes - which is that a new scene doesn't reset the Theme components.

Have you managed to fix this in your WIP PR? I'm not too sure what's the best approach here.

piever:
The cycling palette is somewhat problematic also for grouping because if I start cycling (say color 1, color 2, color 3) and I also have groups but not by color, the color of every group should be color 4, instead by default it keeps cycling, so I at least need to add some code on the grouping side for the two things to work together.

piever:
I've changed the default marker from `Circle` to `:circle` (actually a palette that cycles through some of the possible shapes). At least for the 2D plots the scaling is much nicer and points are always of a reasonable size.

mkborregaard:
What's the reason not to cycle as default?

mkborregaard:
> Have you managed to fix this in your WIP PR? I'm not too sure what's the best approach here.

I have not solved this. I guess the solution is to rebuild the Theme for each new scene.

src/theming.jl
mkborregaard: why does it matter that it is built at precompilation time? Sorry for being slow, but isn't that just a matter of changing the order of includes, or is something happen in precompilation that I'm not realising? Sorry if the question is stupid.

piever: It's a very legitimate question. I simply couldn't find a reshuffling of the "includes" that would allow me to call `Palette(:Dark2)` at this point of the code and didn't want to refactor the code of the package too much. Of course in say MakieThemes or no matter what other package, one would depend on an already precompiled AbstractPlotting and use all the utility functions.

piever:
> What's the reason not to cycle as default?

I though that the user will probably want just one attribute to cycle (most likely color) and not the other, so in `N-1` cases out of `N` you want `cycle=false`. OTOH, cycling color palette may be more common than anything else combined, so I'm not quite sure. Do you have a strong preference either way?

mkborregaard:
> I though that the user will probably want just one attribute to cycle 

That makes sense for predefined palettes, but I think a user providing a palette explicitly would expect it to `cycle`- which seems like an argument for having it `true` by default?

piever:
Makes sense, `cycle = true` is the default now. I've added a resetting mechanism to allow different scenes to coexist without plots on one affecting the cycling palette on the other (fixes https://github.com/JuliaPlots/AbstractPlotting.jl/pull/45#issuecomment-439117494).

piever:
There's actually a very important ingredient that's still missing in the recipe framework wrt cycling palettes: we have no way to tell when to skip or to increase the counter in the palette. For example now `scatterlines` makes the markers and the lines of different colors, even though they should default to the same thing. We need some sort of "primary" concept for a trace.

SimonDanisch:
>scatterlines 

You could likely fix this by fixing the default_theme for scatterlines?

mkborregaard:
I think a new PlotType recipe should always create a PlotType which will constitute one element with the same color, whereas scene recipes can add Plots with different colors

piever:
>> scatterlines
>
> You could likely fix this by fixing the default_theme for scatterlines?

Yes, I've added a method to freeze the theme, just wanted to make sure it actually works as intended in all scenarios.

piever:
It still needs a bit of care to work nicely with grouping, but the overall mechanism should be more or less functional  (with support for "non-primary" series).

piever:
It kind of works now, but the key hack is due from the fact that `convert_attributes` happens only on display, so it's very hard to reason about what to do with grouping, as the user may add some things to the scene, display them, then group other things on top or do the same without displaying in between and these two versions should give the same result.

@SimonDanisch is there a place to have this cycling conversion take place that doesn't depend on whether things are displayed or not? Otherwise, this really requires a lot of acrobatics to work.

SimonDanisch:
>is there a place to have this cycling conversion take place that doesn't depend on whether things are displayed or not?

I was thinking to just have a palette attribute in the scene, and recipes like group should just manually grab their color from those palettes...
E.g.
a sub node a la:

```julia
Scene(palette = (marker = Palette(..), color = Palette(..))))
```

piever:
> I was thinking to just have a palette attribute in the scene, and recipes like group should just manually grab their color from those palettes...

Somehow I think each series should get a "value" and then the scene recipes should combine the value with the palette upon conversion and different recipes decide whether they increase this value or not. What would be a good place to store this value?

piever:
Perfect, so for now I can simply make the palette "non-cylcing" and store it in a `:palette` attribute and we will get the cycling when we also have the "hinting" api.

PR #48, opened 2018-11-18
Mutual exclusive flags/Attributes

louisponet:
This is an attempt to allow one to specify which flags inside a Recipe are mutually exclusive. This is done to avoid conflicts down the pipeline between different attributes. One example of such conflict existed between `:color, :colormap and :colornorm` when trying to plot a `volume`. 

Now what happens is that when merging the flags of the `Theme` or `Scene` with those of the `plot_object`, a check for user specified attributes is done, and the mutually exclusive ones which were defined in the `Theme` or `Scene` are removed.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -518,7 +518,7 @@ convert_attribute(r::AbstractVector{<: Quaternionf0}, k::key"rotation") = r
 
 
 
-convert_attribute(x, k::key"colorrange") = Vec2f0(x)
+convert_attribute(x, k::key"colorrange") = x==nothing ? nothing : Vec2f0(x)
 
 convert_attribute(x, k::key"textsize") = Float32(x)
 convert_attribute(x::AbstractVector{T}, k::key"textsize") where T <: Number = el32convert(x)
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -22,6 +22,11 @@ function default_theme(scene)
     )
 end
 
+
+#this defines which attributes in a theme should be removed if another attribute is defined by the user,
+#to avoid conflicts later through the pipeline
+
+mutual_exclusive_attributes(::Type{<:AbstractPlot}) = Dict()
 """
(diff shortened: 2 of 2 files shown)


mschauer:
`mutually_exclusive` suggest symmetry, but the action removes one attribute in presence of a second, I'd call that `superseeded_by` or something along those lines.

louisponet:
MWE:
```
using Makie
rs = range(-1, stop=1, length=100)
dens = [x^2 + y^2 + z^2 for x=rs,y=rs,z=rs]
volume(dens, color=:blue, algorithm=:iso)
```

louisponet:
That makes more sense indeed. or something like `overrides`, since it could also be the other way around when the user specifies `colormap`. Right now there is some sense of hierarchy created by the `_default` functions in `GLVisualize` where for example there is a definition `color = color_map == nothing ? default(RGBA, s) : nothing`. Which is also why for now I'm not deleting the flags but rather putting them to `nothing` during the check.

louisponet:
So about the symmetry: I would say it should be there, it depends on what the user specifies, however the current system with `_default` did not allow that, which is why for now I didn't work on the vice versa situation.

louisponet:
I retracted my previous comments. I didn't think it through, this way I think it's more symmetric and also errors when people try to specify mutually exclusive flags at the same time. 

BTW I don't know if conflicts like this exist in other parts of the code, I just ran into it using `volume` so maybe some system like this is not necessary.

mkborregaard:
So if I specify `colormap` and `color` in a Theme, which will have priority when defining the color of a Plot? I'd expect `color`.

SebastianM-C:
With `color` and `colormap` I think you can use both if you give numerical values to `color` and the actual colors to `colormap`.

mkborregaard:
Yes, but that doesn't work in a Theme - where I want to pass a color or a `Palette` to `color`, and a gradient to `colormap`.

PR #52, opened 2018-11-27
widen function recipe signature

piever:
I thought for consistency with other recipes, this should actually dispatch on `PlotFunc` rather than `Type{<:AbstractPlot}`


piever:
I've also added `Lines` as default plot type, as I think it is by far the most natural choice to plot a function of one variable.

PR #54, opened 2018-11-28
add default_palettes

piever:
This adds palettes to the theme, as proposed in https://github.com/JuliaPlots/AbstractPlotting.jl/pull/45#issuecomment-442474327. We dodge the problem of cycling (to be implemented later) but overall this is a much cleaner implementation than #45

I'm moving StatsMakie default scales here so that they can be easily themable. In the future we can thing of which other attributes could have a default palette.

### DIFF
--- a/src/basic_recipes/multiple.jl
+++ b/src/basic_recipes/multiple.jl
@@ -40,14 +40,14 @@ function convert_arguments(P::PlotFunc, m::PlotList)
     PlotSpec{MultiplePlot}(pl)
 end
 
-combine(o1, o2, o3) = o3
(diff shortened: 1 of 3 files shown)


piever:
I'm taking advantage of this PR to also make more things themable (marker, markersize, linestyle), though this is probably not crucial and if it is problematic I can take it out.

SimonDanisch:
Cool :) we'll need to be a bit careful with globals like that, but we can leave that for another day! I don't really know what the best solution will be anyhow (for static compilation) ;)

daschw:
Awesome, I'm glad to see the wong colors.

PR #55, opened 2018-11-30
RFC: linesfill and ribbon

piever:
A possible implementation: still do decide whether we want it as a recipe or an attribute.


piever:
Closing as `fill` makes more sense as a `Lines` attribute and `Ribbon` belong to StatsMakie as a way to turn a mean + error into `Lines` + `Band`

PR #56, opened 2018-11-30
define product on Combined

piever:
In the spirit of ggplot2, here is a simple way to define product on `Combined`, so that the user con put a few of those together. I find it useful because it feels a bit artificial to define things like `scatterlines` to have both lines and scatter already. Here for example one could do:

```julia
plot(Scatter*Lines, -pi..pi, sin)
```
screenshot from 2018-11-30 15-34-23

or:
```julia
julia> plot(BarPlot*Lines, histogram(nbins = 20), linewidth = 3, randn(1000))
```
screenshot from 2018-11-30 15-34-36

It's actually useful for performance in StatsMakie because I can do the grouping of the arguments only once.

### DIFF
--- a/src/basic_recipes/multiple.jl
+++ b/src/basic_recipes/multiple.jl
@@ -66,7 +66,7 @@ function plot!(p::Combined{multipleplot, <:Tuple{PlotList}})
 end
 
 function default_theme(scene, ::Type{<:Combined{S, T}}) where {S<:Tuple, T}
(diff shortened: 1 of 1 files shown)


piever:
One extra idea would be to allow the various series to have private arguments and keyword arguments. For example:

```julia
plot(Lines(smooth; color = :red)*Scatter, Data(iris), :SepalLength, :SepalWidth)
```

would add a smoothed version of the scatter as a line in red.

PR #60, opened 2018-12-13
add labels + axis

SimonDanisch:
image


src/basic_recipes/basic_recipes.jl
```diff
@@ -419,6 +419,11 @@ convert_arguments(::Type{<: BarPlot}, y::AbstractVector{<: Number}) = (1:length(
 function AbstractPlotting.plot!(p::BarPlot)
     pos_scale = lift(p[1], p[2], p[:fillto], p[:width]) do x, y, fillto, hw
         nx, ny = length(x), length(y)
+        if x isa AbstractVector{<: AbstractString}
```
piever: Plots also support scatter plots with categorical variables, so maybe better to support things there? Say `scatter(["a", "b", "c"], ["d", "e", "f"])`

SimonDanisch: where? you mean in general?

piever: Yes, given that we may wish to support scatter plots with discrete variables and barplot would call scatter, I thought it made more sense to fix this in scatter.

As a separate point `AbstractString` is maybe not the only thing we wish to support (symbols should work too I guess, and maybe other things in future, say dates or unitful axes), so it may be better to do this with a trait, to make the mechanism extensible.

SimonDanisch: so the general transform seem to be an order preserving unique and that's then mapped to something like 1:n?

SimonDanisch: I mean, in theory you could also have meshes/images and stuff like that... anything displayable really

piever: I would think so, some sort of `iscategorical` trait that would tell whether to do the transform and then map to `1:n` while preserving the ticks, or adding a general method to convert the type to tick, say the compact format like [here](https://github.com/piever/InteractBase.jl/blob/master/src/slider.jl#L6) (we want to make sure it also works for strings from a `CategoricalArray` which are wrapped in a `Categorical` type).

PR #63, opened 2018-12-22
fix `on(f, camera, node)`

Datseris:
having an additional docstring *without* the function call signature is extremely confusing.


SimonDanisch:
thanks!

PR #67, opened 2019-01-15
bump version to reflect tag

colinxs:
Makie.jl requires >= 0.9.2 of AbstractPlotting but the tagged version (and master) don't declare 0.9.2


SimonDanisch:
thanks!

PR #74, opened 2019-01-24
flip the colorbar to make it coherent with the colormap scale

alblaz:
The current implementation of `colorlegend` results in inverted colorbars. As an example the function exp(-(x^2 + y^2)) yields the following meshplot
testmesh

This pr fixes this problem as shown in the figure below
testmesh_fixed


SimonDanisch:
thank you!

PR #79, opened 2019-02-17
Add a `framerate` optional argument to all recording

asinghvi17:
`ffmpeg` only functions correctly when the input and output framerates are the same by [this bug report](https://trac.ffmpeg.org/ticket/3856), so the `framerate` option, type set to `Int` and default value 24, is passed to both the `VideoStream` and `save` functions.

~Please note that this PR is by no means final, and I haven't done testing on it yet; I'm just putting it out there for feedback.~

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -206,14 +206,15 @@ end
 
 
 """
-    VideoStream(scene::Scene, dir = mktempdir(), name = "video")
+    VideoStream(scene::Scene, dir = mktempdir(), name = "video"; fps = 24)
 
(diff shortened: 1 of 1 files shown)


asinghvi17:
Works on my machine.  @SimonDanisch would you like to test as well?

asinghvi17:
This PR should fulfill JuliaPlots/Makie.jl#287 as well, @Datseris

SimonDanisch:
Thanks a lot! This is ready to merge, right?

asinghvi17:
As far as I can test it on my machine :)

PR #82, opened 2019-02-19
Export a method `save` from AbstractPlotting

asinghvi17:
This is meant to (a) bring AbstractPlotting in line with the documentation and (b) make it simpler to save.  As of now, we have to import FileIO in order to save.  However, it is much simpler in my opinion to simply export a save function from AbstractPlotting so you don't have to load any extraneous packages.


asinghvi17:
Should fix JuliaPlots/Makie.jl#293

SimonDanisch:
Thanks! I think it'd be better to just reexport FileIO.save?!

SimonDanisch:
Thanks for the quick PR, but I found a couple of other things I want to fix related to `save`, so it's best if I just fix it!

PR #84, opened 2019-02-22
Print scientific notation in `slider`

asinghvi17:
Currently, the status is that the `default_printer` will print in normal numbers but revert to scientific notation if the number gets too small, to fix JuliaPlots/Makie.jl#241.  However, I would also like to implement a `sci_printer` to print scientific notation always, though I haven't quite figured out how to do that with round (hence the draft PR.

### DIFF
--- a/src/AbstractPlotting.jl
+++ b/src/AbstractPlotting.jl
@@ -115,7 +115,6 @@ export disconnect!, must_update, force_update!
 
 # gui
 export slider, button, playbutton
-export default_printer, sig_printer
 
 # Raymarching algorithms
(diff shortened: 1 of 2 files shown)


asinghvi17:
I am now going to add a dependency on `Printf` to use the `@sprintf` macro.  This will allow consistent scientific notation on both large and small numbers.

asinghvi17:
All that remains now IMO is testing and documentation.   @SimonDanisch would you mind giving some feedback on if I can do anything better to the code?

SimonDanisch:
Cool thank you :) that looks good to me!

PR #88, opened 2019-02-27
widgets integration

piever:
Heavily WIP, the plan is to allow:

```julia
@manipulate for i=1:100, j=1:100
    scatter(rand(i), rand(i), markersize = j)
end
```

to work.

There is one thing I'm not sure how I do: the `Widget` type here can render like a scene (in that `Widget.slider(1:100)` displays in a Makie window), but I'm not sure how can I get `hbox(Widget.slider(1:100), Widget.slider(1:100))` to work...

The "type piracy" lines on `Widget` should be moved to Widgets.jl of course (`show`, `display` and `render` methods).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -27,6 +27,7 @@ SignedDistanceFields = "73760f76-fbc4-59ce-8f25-708e95d2df96"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 UnicodeFun = "1cfade01-22cf-5700-b092-accc4b62d6e1"
+Widgets = "cc8bc4a8-27d6-5769-a93b-9d913e69aa62"
 
 [compat]
 GeometryTypes = ">=0.7.2"
--- a/Project.toml
+++ b/Project.toml
@@ -31,6 +31,7 @@ Widgets = "cc8bc4a8-27d6-5769-a93b-9d913e69aa62"
 
 [compat]
 GeometryTypes = ">=0.7.2"
+Widgets = ">=0.5.1"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
--- a/REQUIRE
+++ b/REQUIRE
@@ -18,4 +18,4 @@ FileIO
 FixedPointNumbers
 Contour
 ImageMagick
-Widgets 0.5.0
+Widgets 0.5.1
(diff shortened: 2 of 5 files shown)


piever:
`@manipulate` is almost working (needs https://github.com/piever/Widgets.jl/pull/28), with the only issue that I'm not sure how to convert an `Observable{Scene}` (which, I agree, is not the most performant way to do interactivity but it is Interact defaults) into something that can be plotted with Makie.

piever:
I'm getting somewhere (with this PR plus https://github.com/piever/Widgets.jl/pull/29). Preview video available [here](https://vimeo.com/323275432).

The basic idea is that `@manipulate` is actually a very poor fit for Makie (as it would try to redraw the whole plot every time a widget gets changed). A better solution is to use the newish macro `Widgets.@auto` that converts the same values you'd pass to manipulate to widgets with the correct label and then feed those observables (or transformations of them) directly to the plot.

Example:

```julia
julia> using Widgets, AbstractPlotting, GLMakie, Distributions

julia> Widgets.set_backend!(AbstractPlotting.MakieBackend());

julia> wdgs = @auto μ = -10:10 σ = 1:10;

julia> x = range(-10, 10, length = 100);

julia> y = map((a, b) -> pdf.(Normal(a, b), x), μ, σ);

julia> Widget(wdgs, output = lines(x, y))
```

The identical syntax would work if one uses the Interact backend. Unfortunately only `slider` is wrapped at the moment so this "manipulate like" thing only works for ranges (I'd need to wrap also textboxes, colorpickers, togglebuttons etc. for it to work more generally).

SimonDanisch:
Should we merge it, even though it's kind of incomplete?

piever:
There are 3 open issues I think:

1. What is the correct fallback for `vbox` and `hbox`? Here I'm afraid I'm abusing `convert`, maybe we should check if it is a `Transformable` and otherwise call `Scene` on it, or maybe we can use something liker `AbstractPlotting.render` and call it before applying the layout (to turn an object into a `Transformable`)
2. Should the Widgets version replace the native version? Meaning, should the AbstractPlotting `slider` be renamed to something like `rawslider` and we encourage user to use the Widgets version
3. How do we render an `Observable{Scene}`? Meaning, we need some way of saying that if the whole `Scene` is replaced by something else, we should redraw the whole thing. This is not the performant way to update things but may happen / be necessary.

I would like to figure out 1. before merging (I think 2. and 3. can wait). What's your view on point 1?

asinghvi17:
Bump @piever @SimonDanisch

asinghvi17:
Is this working as intended as of now?  (Aside from missing interaction features)

piever:
There maybe some redesign of Widgets, better hold off this one for now.

PR #91, opened 2019-03-03
add unit support [WIP]

SimonDanisch:
needs propper integration with backends & tests

### DIFF
--- a/src/basic_recipes/basic_recipes.jl
+++ b/src/basic_recipes/basic_recipes.jl
@@ -522,8 +522,7 @@ Creates a contour plot of the plane spanning x::Vector, y::Vector, z::Matrix
     pop!(default, :color)
     Theme(;
         default...,
-        color = nothing,
-        colormap = theme(scene, :colormap),
+        color = theme(scene, :colormap),
         colorrange = AbstractPlotting.automatic,
         levels = 5,
         linewidth = 1.0,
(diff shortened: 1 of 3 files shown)


c42f:
Cool, looks interesting Simon. A couple of questions:

* Did you consider using Unitful.jl ?
* What are the contour changes here, is that just cleanup?

What do you think about encouraging DIP as a psuedo angular-like unit for use by default?  As I wrote at #225, I feel that the feature size as an *angle at the user's eye* is the thing which actually matters in typical usage, not the physical size of the screen in mm.

Generic recipes should thus not care about mm! They care about angles.

SimonDanisch:
I was writing it with the idea in mind to have DIP the default ;) I've been bitten by mm too much^^ It's simply impossible to have it scale nicely on all different screens.

>Did you consider using Unitful.jl ?

I think I did - I don't remember anymore why I decided against this (this code is pretty old, this PR just udpated it for the new Makie^^)

>What are the contour changes here, is that just cleanup?

Woops, shouldn't be part of this PR

c42f:
I think we're on the same page :-)

Regarding Unitful, I think it's pretty nice, my only reservation might be that it's a little complex for incidental use. It did previously have a problem with defining units outside of `Unitful` itself and referring to them using the `@u_str` macro, but I fixed that a few months ago.

ssfrr:
:+1: for using Unitful if possible. If your code already has `using Unitful: mm` it's nice not to have conflicting identifiers when you want to use the same units for plotting.

c42f:
Personally I've been tending to use the `u_str` macro to avoid such conflicts. It's not as pretty though.

phlavenk:
I'm definitely supportive for easy interaction with `Unitful`. It is a wonderful package, but I do agree that AbstractPlotting should not need to depend on it.
It would be great if there would be a lightweight `AbstractUnits.jl` package that would define a minimal API so that different unit handling packages can exchange data. However, I don't see any easy way how to do the conversion between apriori unknown types, even only for length in the context of a graphical layout. Maybe you could ask @ajkeller34 for his opinion.

mkborregaard:
Would be nice and simple to just depend on Unitful. But the package has a load time of 1.5 seconds on my system.

phlavenk:
Yeah, that's too long. 
Therefore, it would be great to have only an abstract API package - like, extract the type definition from Unitful and use this very same type through AbstractPlotting. And re-implement a private function `convert_to_mm_units(a::AbstractUnitful.Length)` defined in AbstractPlotting.
Otherwise there would be a need for AbstractPlottingUnitfulUtils just to be able to write `mm` in the code...

c42f:
Well, it's worth keep in mind that `using AbstractPlotting` takes several times that long, and it's not clear to me how that can be solved without native code precompliation (which would also solve the problem for `Unitful`).

Lightweight integration would be great of course if it's achievable.

Also possibly relevant (though attacking a different part of the problem): https://github.com/ajkeller34/UnitfulPlots.jl

asinghvi17:
Seems like that's overloading the axis number formatter.  That would be useful to have, actually - I'll try to add a formatter in to the axis refactoring I'm working on.

asinghvi17:
What is remaining to do here?

(of course, these all need to be documented, but that's easy)

asinghvi17:
Should we close this in favour of #185?

PR #96, opened 2019-03-14
Add PlotUtils color gradient support

asinghvi17:
This PR aims to:

- [x] Integrate PlotUtils' colormaps into Makie

- [x] Remove `viridis` special-case handling

- [x] Remove ColorBrewer special case handling (except for 8-color only palettes)

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -699,11 +699,8 @@ function convert_attribute(cs::Union{String, Symbol}, ::key"colormap", n::Intege
     if cs_string in all_gradient_names
         if cs_string in colorbrewer_8color_names
             return resample(ColorBrewer.palette(cs_string, 8), n)
-        else
-            return resample(ColorBrewer.palette(cs_string, 9), n)
-        end
-    elseif cs_string in plotutils_names
-        return PlotUtils.cvec(Symbol(cs), n)
+        else # cs_string in plotutils_names
+            return PlotUtils.cvec(Symbol(cs), n) .|> color .|> x -> convert(RGB{FixedPointNumbers.Normed{UInt8,8}}, x)
         end
     else
         error("There is no color gradient named: $cs")
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -663,7 +663,7 @@ function available_gradients()
 end
 
 """
-Reverses the attribute T uppon conversion
+Reverses the attribute T upon conversion
 """
 struct Reverse{T}
     data::T
@@ -698,9 +698,11 @@ function convert_attribute(cs::Union{String, Symbol}, ::key"colormap", n::Intege
     cs_string = string(cs)
 
     if cs_string in all_gradient_names
-        if cs_string in colorbrewer_8color_names
+        if cs_string in colorbrewer_8color_names # special handling for 8 color only
             return resample(ColorBrewer.palette(cs_string, 8), n)
-        else # cs_string must be in plotutils_names
+        elseif cs_string ∈ colorbrewer_names    # get colormaps directly from ColorBrewer, prefer that to PlotUtils
+            return resample(ColorBrewer.palette(cs_string, 9), n)
+        else                                    # cs_string must be in plotutils_names
             return PlotUtils.cvec(Symbol(cs), n) .|> color .|> x -> convert(RGB{FixedPointNumbers.Normed{UInt8,8}}, x)
         end
     else
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -647,7 +647,7 @@ to_colormap(x::Union{String, Symbol}, n::Integer) = convert_attribute(x, key"col
 
 """
 A Symbol/String naming the gradient. For more on what names are available please see: `available_gradients()`.
-For now, we support gradients from `PlotUtils` and `ColorBrewer` natively.
+For now, we support gradients from `PlotUtils` natively.
 """
 function convert_attribute(cs::Union{String, Symbol}, ::key"colormap", n::Integer = 20)
     cs_string = string(cs)
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -612,7 +612,7 @@ Prints all available gradient names.
 """
 function available_gradients()
     println("Gradient Symbol/Strings:")
-    for name in sort(collect(@eval PlotUtils.cgradients()))
+    for name in sort(collect(all_gradient_names))
         println("    ", name)
     end
 end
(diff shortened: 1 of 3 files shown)


asinghvi17:
This is really weird.  After this particular commit, 
```julia
using Makie
lines(1:10, 1:10, color = 1:10, colormap = :magma)
```
gives an empty plot, but 
```julia
using PlotUtils, FixedPointNumbers
cs = :magma
n = 20
lines(1:10, 1:10, color = 1:10, colormap = PlotUtils.cvec(Symbol(cs), n) .|> color .|> x -> convert(RGB{FixedPointNumbers.Normed{UInt8,8}}, x), linewidth = 10)
```

works fine.  The colormap here should be the same as the colormap there - I literally copy-pasted the handling code for it.  Is there something I'm doing wrong?  (The code is being eval'ed using Atom, that's how I'm testing it).

mkborregaard:
string/symbol thing?
BTW the diff looks weird, why is that?

asinghvi17:
@mkborregaard diff should be fixed now. And the problem seems to be solved as well, testing in Atom was a pain.  I ended up just removing my PackageCompiler sysimage for testing.

asinghvi17:
Also, this PR should be ready to merge now.  This will break the MakieGallery example with all the colorschemes (because there are so many more now).

asinghvi17:
See also JuliaPlots/MakieGallery.jl#28 - to be merged only after the next tag of AbstractPlotting.

src/conversions.jl
```diff
@@ -593,59 +593,62 @@ convert_attribute(x, k::key"linewidth") = Float32(x)
 convert_attribute(x::AbstractVector, k::key"linewidth") = el32convert(x)
 
 const colorbrewer_names = Symbol.([
-    # All sequential color schemes can have between 3 and 9 colors. The available sequential color schemes are:
```
mkborregaard: This diff is still strange

src/conversions.jl
```diff
@@ -688,43 +691,21 @@ end
 to_colormap(x::Union{String, Symbol}, n::Integer) = convert_attribute(x, key"colormap"(), n)
 
 """
-A Symbol/String naming the gradient. For more on what names are available please see: `available_gradients()
+A Symbol/String naming the gradient. For more on what names are available please see: `available_gradients()`.
+For now, we support gradients from `PlotUtils` and `ColorBrewer` natively.
```
mkborregaard: ColorBrewer gradients are also in PlotUtils. No reason to keep them here too.
Also, I think it would be better to list the color libraries that are available, as people will know those (and because of licensing)

mkborregaard:
Sorry, I approved a little too fast, still have some questions and concerns.

First of all, colorbrewer gradients are now included twice, because they are also in PlotUtils.
Also, the color gradients of PlotUtils are in different libraries. That's for convenience and resolving naming conflicts. We could make that available here by reexporting `clibrary` to let users choose the current library. 
There are also tools to access gradients from them command line. 
Does `available_gradients()` show the PlotUtils gradients now as well? Otherwise PlotUtils has `cgradients` which will give all gradients in the current library.

Finally, most gradients are on licenses that require attribution. That's why I think the libraries should be transparent to the user, and mentioned in the docs somewhere, if briefly.

src/conversions.jl
asinghvi17: PlotUtils doesn't have all ColorBrewer gradients unfortunately...if you try `AbstractPlotting.colorbrewer_names in AbstractPlotting.plotutils_names` it returns false.

asinghvi17:
`available_gradients()` should show the PlotUtils gradients as well, because it is just a wrapper around `all_gradient_names`.  If I need to add `clibrary` functionality (which makes sense now that you mention it), then we would have to convert `all_gradient_names` and `plotutils_names` to functions which check the available colour library when run.  Is there a better way to do this besides `@eval PlotUtils._gradients[1]`?

mkborregaard:
About the colorbrewer gradients - I think the ones missing in PlotUtils are those that are not gradients, but discrete collections of colors. These should be Palette's, not gradients - they are useful for automatically shifting colors with new series (indeed the default palette is one of the discrete colorbrewer ones), rather than showing magnitudes of numbers (which is what the sequential and diverging ones do).

About using the clibrary functionality, that is already in PlotUtils, so shouldn't you just be able to use PlotUtils' accessor functions to get them?
If you've got suggestions to change that system that would also be fine.

asinghvi17:
Ah, didn't realize about the palette part.  That should be fairly easy to clean up, then - I'll have a more in depth look at the PlotUtils API once I'm off my flight.  Hopefully this will make the code a lot cleaner.

SimonDanisch:
So this isn't good to merge then?!

asinghvi17:
So PlotUtils' `clibrary` function only changes handling of namespace clashes - that is, if there are multiple colormaps (provided by different libraries) which share the same name - otherwise, it seems it goes with some default handling.  Since PlotUtils handles this natively, I think we should be good to go on this.  It's simply a matter of the user knowing what to do, which I think JuliaPlots/MakieGallery.jl#28 should address.

asinghvi17:
@SimonDanisch @mkborregaard this should finally be ready.  The functionality was already mostly in PlotUtils, this just involved converting the PlotUtils output from `RGBA{Float32}` to `RGB{N0f8}` (and removing Colorbrewer special case handling thanks to mkborregaard's info.

src/conversions.jl
```diff
@@ -654,13 +612,13 @@ Prints all available gradient names.
 """
 function available_gradients()
     println("Gradient Symbol/Strings:")
-    for name in sort(collect(all_gradient_names))
+    for name in sort(collect(@eval PlotUtils.cgradients()))
```
SimonDanisch: eval!?

mkborregaard: Should be no reason for that, the function just returns a vector of symbols

asinghvi17: Woops, forgot to push cleanup...will do that when I get back to my laptop.

mkborregaard: I'm a bit embarrassed I missed that

asinghvi17: I'd put that there as a test, it was removed in the cleanup commotion that I havent pushed yet ;)

SimonDanisch:
I'm not sure why I didn't see the call, that this is actually ready ;) Sorry, and thanks a lot for doing this!

asinghvi17:
My pleasure :) .  This PR was a bit of a mess because between spring break and midterms I didn't get much time to work on it, so there was a bit of confusion on my part as well :P

PR #98, opened 2019-03-18
Correct axes bug for heatmaps & images

ianshmean:
As discussed in https://github.com/JuliaPlots/Makie.jl/issues/295 and as a secondary point in https://github.com/JuliaPlots/Makie.jl/issues/205

For this example:
```
image(Makie.logo(), scale_plot = false)
```
Master gives:
image

This PR gives:
image

I'm late to the discussion, but my view is that for images, I do think there needs to be `transpose` and `flipyaxis` options that both should default to `true` for use of `image`.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -235,7 +235,7 @@ and stores the `ClosedInterval` to `n` and `m`, plus the original matrix in a Tu
 """
 function convert_arguments(::SurfaceLike, data::AbstractMatrix)
     n, m = Float32.(size(data))
-    (0f0 .. m, 0f0 .. n, el32convert(data))
+    (0f0 .. n, 0f0 .. m, el32convert(data))
 end
 
 """


ssfrr:
does this PR need to be a WIP or is it ready to merge?

daschw:
```julia
Plots.plot(Makie.logo(), aspect_ratio = true)
```
gives
Makie_logo
which I think makes more sense, no?

SimonDanisch:
Yeeeeah, so this is a bit complicated, if we want to make everyone happy...
I think I decided, that `image(img)` should do what you want, and `heatmap(img)` should work, but should be displayed like in this PR.

ianshmean:
Just a thought.. Do people find the meaning of `scale_plot = false` to be clear? I'm wondering whether `squarepixels = true` would be clearer, where true would be default?

ssfrr:
> Just a thought.. Do people find the meaning of scale_plot = false to be clear? I'm not sure that I do, and I'm wondering whether squarepixels = true would be clearer, where true would be default?

maybe worth a separate issue? I know we discussed this on slack a while ago and some alternates came up that I think @SimonDanisch was OK with.

ssfrr:
bump. this seems like a strict improvement/bugfix. I often plot non-square heatmaps so this would be helpful to me.

PR #99, opened 2019-03-19
Mark inplace vbox with !

JonasIsensee:
I found this while looking through the code.
It seem that this function operates inplace and the `!` was left out on accident.
The corresponding `hbox!` function defined below it, has it.


asinghvi17:
Pinging @SimonDanisch - seems good to me...

SimonDanisch:
Thank you!

PR #106, opened 2019-04-02
[Urgent] Fix colors being converted to colormaps

### LINKED ISSUE #105: Error with colours after PlotUtils integration
From Slack (@miguelraz)

It seems that PlotUtils integration has hijacked something to do with the color conversion pipeline.

```julia
julia> vbox(
        hbox(s1, s2, s3, s4, s5, s6),
        mesh_scene, parent = parent)
┌ Warning: gray is found in more than one library: cmocean, colorcet. Choosing cmocean
└ @ PlotUtils ~/.julia/packages/PlotUtils/GxT73/src/color_gradients.jl:40
Error showing value of type Scene:
ERROR: MethodError: no method matching color(::Array{RGB{FixedPointNumbers.Normed{UInt8,8}},1})
```

I think this should probably just return a single colour, so no clue why it's doing that.

It could also be that to_colormap is overriding to_color...

asinghvi17:
Colors were being converted to colormaps seemingly arbitrarily in here:

https://github.com/JuliaPlots/AbstractPlotting.jl/blob/f900f0befce4bd059b7aa795f091e7af5b1bc4ad/src/conversions.jl#L446-L449

With the PlotUtils integration, many seemingly innocent names like `gray` were being parsed as Colormaps.

This PR fixes that and that alone.  Since this bug is breaking master at the moment, it should probably get looked at ASAP.

cc: @mkborregaard @miguelraz

(Simon is on vacation so didn't want to bother him too much :P)

Fixes #105.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -443,10 +443,7 @@ end
 
 convert_attribute(c::Colorant, ::key"color") = convert(RGBA{Float32}, c)
 convert_attribute(c::Symbol, k::key"color") = convert_attribute(string(c), k)
-function convert_attribute(c::String, ::key"color")
-    c in all_gradient_names && return to_colormap(c)
-    parse(RGBA{Float32}, c)
-end
+convert_attribute(c::String, ::key"color") = parse(RGBA{Float32}, c)
 
 # Do we really need all colors to be RGBAf0?!
 convert_attribute(c::AbstractArray{<: Colorant}, k::key"color") = el32convert(c)


mkborregaard:
Would it not make more sense to insure that there are no colormap names that overlap with color names? Might be a change in PlotUtils. I don't see why `:gray` would be in `all_gradient_names`?

asinghvi17:
`:gray` is from both `colorcet` and `cmocean`, and if we removed them from PlotUtils it'd probably break a lot of code in Plots...

asinghvi17:
We could check for the existence of the color in Colors.jl's color names before going through PlotUtils, which would probably be the better solution

mkborregaard:
I just renamed them in PlotUtils. Not too worried about breaking that. Is this a problem for other colors or is the problem solved with this?

https://github.com/JuliaPlots/PlotUtils.jl/commit/bec5632d6d1e851c8b482910dad4cb32cb89d19d

asinghvi17:
I think this is the only place this pops up, but I'll run MakieGallery on AbstractPlotting master to make sure.  Thanks!

asinghvi17:
Yep, it works now - looks like `:gray` was an edge case.

mkborregaard:
Awesome, thanks for acting on this so quickly.

PR #107, opened 2019-04-02
[Feature] Docstrings in @recipe transfer to plotting functions

asinghvi17:


asinghvi17:
[See this macro](https://docs.julialang.org/en/v1/manual/documentation/index.html#Core.@__doc__), it seems to be exactly what we're looking for here...this will probably get rid of the boilerplate macros, but we could document that somewhere else.

mkborregaard:
For Plots this type of feature was implemented in https://github.com/JuliaPlots/RecipesBase.jl/pull/34

asinghvi17:
Lovely, thanks!  That was what I was thinking of doing as well - and this actually solves the problem of evaluating `@__doc__` at the proper time!

asinghvi17:
Should be ready for review, changes are pretty minor.  Note that we will need to provide docstrings for all recipes now, since they won't have the boilerplate ones, but that should be fairly routine.

PR #108, opened 2019-04-03
[WIP] Document recipes, add CI

asinghvi17:
With #107, recipe docstrings will transfer to their generated functions.  To make the most of this, all recipes in AbstractPlotting should, in my opinion, be documented. 

For this, we'll have to go through the entire codebase and see what each `@recipe` does and what the argument conversion pipelines work like.  A lot of things are also completely undocumented even in JuliaPlots/MakieGallery.jl, and we will probably need to upload examples there as well.

Below is the list of recipes to be documented.  Please feel free to help out by posting what those recipes do, or giving an example of your usecase to MakieGallery.

- [x] arrows
- [ ] StreamLines
- [ ] Series (is this needed or superceded by PlotList?)
- [x] Arc
- [x] BarPlot
- [ ] ScatterLines
- [x] Band
- [ ] VolumeSlices
- [x] Legend
- [ ] Slider
- [ ] Button

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ prec.zip
 tmax.zip
 bluemarble-2048.png
 phases.0001_print.jpg
+**/._*
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,27 +2,28 @@
 language: julia
 os:
   - linux
-  - osx
 julia:
-  - 0.6
+  - 1.1
   - nightly
 notifications:
   email: false
 git:
   depth: 99999999
+env:
+  - IS_TRAVIS_CI = "true"
 
-## uncomment the following lines to allow failures on nightly julia
-## (tests will run but not make your overall status red)
-#matrix:
-#  allow_failures:
-#  - julia: nightly
+# uncomment the following lines to allow failures on nightly julia
+# (tests will run but not make your overall status red)
+matrix:
+ allow_failures:
+ - julia: nightly
 
 ## uncomment and modify the following lines to manually install system packages
 #addons:
-#  apt: # apt-get for linux
-#    packages:
-#    - gfortran
-#before_script: # homebrew for mac
+ apt: # apt-get for linux
+   packages:
+   - ffmpeg
+# before_script: # homebrew for mac
 #  - if [ $TRAVIS_OS_NAME = osx ]; then brew install gcc; fi
 
 ## uncomment the following lines to override the default test script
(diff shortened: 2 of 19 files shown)


asinghvi17:
Just a little explanation here - I merged my experimental AbstractPlotting CI into this so I could test for errors internally.

asinghvi17:
Since I've merged CI testing in here as well, I might as well make this a joint CI + doc commit.  

Initial result of first trial test is [here](https://travis-ci.com/asinghvi17/AbstractPlotting.jl/jobs/190818457)

asinghvi17:
I think the tests are in a pretty good state now, see the output [here](https://travis-ci.com/asinghvi17/AbstractPlotting.jl/jobs/190831777).  It needs to be cleaned up of course, but I think it's probably a decent way to at least test if something broke.  Takes about a half hour per run, but that's par for the course I'm afraid (unless we add specific indices to test)...

asinghvi17:
Test/REQUIRE was failing to work when I tried it, hence extras - also, they're panning to move to a test/Project.toml setup in v1
2 so that should be an easier copy + paste then :)

SimonDanisch:
Hm, I did start using that as well, but run into pretty weird behavior, so I decided to use REQUIRES until pkg has officially deprecated the REQUIRE file ;)

asinghvi17:
I believe 
```toml
[extras]
Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
[targets]
test = ["Test"]
```
is still required to run tests on Pkg...added that.  See the error log of [this Travis CI build](https://travis-ci.org/JuliaPlots/AbstractPlotting.jl/builds/517161773?utm_source=github_status&utm_medium=notification).

asinghvi17:
What was the weird behaviour you were seeing there?  The problem here on Travis is that the requisite packages aren't being installed by `test/REQUIRE`- either Pkg has already deprecated them or something odd is going on, but I don't get this error on Travis using `Project.toml`.
```julia

   Testing AbstractPlotting
 Resolving package versions...
ERROR: LoadError: ArgumentError: Package MakieGallery not found in current path:
- Run `import Pkg; Pkg.add("MakieGallery")` to install the MakieGallery package.
```

asinghvi17:
Also, inspired by GR's documentation, would it make sense to list parameters in a table in our docstrings as well?  Something like:
```julia
help?>lines
```
`lines(x, y, z)` / `lines(x, y)` / or `lines(positions)`

  Creates a connected line plot for each element in (x, y, z), (x, y) or positions.

Attributes:
[...]

asinghvi17:
All functionality from this branch has been merged via #112, closing.

Issue #109, opened 2019-04-03
UndefVarError: removehandler_callbacks not defined


goretkin:
At https://github.com/JuliaPlots/AbstractPlotting.jl/blob/e2f48d328b3341da367bd41611df31627bb19f6d/src/interaction/nodes.jl#L53

Seems to come from a missing definition. It can probably be removed altogether.
https://github.com/JuliaGizmos/Observables.jl/blob/77e984fc72765ab1b30486809300255c1361f715/src/Observables.jl#L8

This never gets triggered, except when I was trying to use `@enter` from `Debugger.jl`

asinghvi17:
Fixed by #111

PR #111, opened 2019-04-05
Remove undefined reference

goretkin:
Fixes issue #109


SimonDanisch:
This should be `Observables.removehandler_callbacks` ;)

goretkin:
Thought that might be the case :)

SimonDanisch:
thank you!

PR #112, opened 2019-04-09
test with MakieGallery

SimonDanisch:


asinghvi17:
Shall we also use Gitlab CI within AbstractPlotting.jl to test rendering?  Perhaps an environment variable could be set?

asinghvi17:
Also, I think you might have to regenerate the reference images.

asinghvi17:
Should we merge this now?  Seems like it works.

SimonDanisch:
Thank you :)

PR #114, opened 2019-04-11
Button tweaks

grero:
As I was playing around with the buttons part of the GUI tools, I noticed a few short comings. This is my attempts at rectifying some of them.

### DIFF
--- a/src/interaction/gui.jl
+++ b/src/interaction/gui.jl
@@ -182,7 +182,7 @@ function plot!(splot::Button)
         splot, txt,
         color = textcolor,
         textsize = textsize, position = position,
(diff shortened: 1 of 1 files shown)


asinghvi17:
@grero would you say this is finished?

grero:
Hm, I'd say it implements want I needed for my particular use, but I defer to you experts whether it's worth merging in its current state : )

asinghvi17:
Well, if there aren't any other changes you'd like to make, we may as well - and more pull requests can always be made for new features :)

SimonDanisch:
Thank you, that looks good :)

PR #119, opened 2019-04-18
title improvements

SimonDanisch:


asinghvi17:
In that case, would it be better to call the function `title`?  For consistency with Plots and all.

SimonDanisch:
yeah, I just stepped over that as well

asinghvi17:
Documentation in JuliaPlots/MakieGallery.jl#44

PR #124, opened 2019-05-03
Make MP4s have better quality

asinghvi17:


asinghvi17:
Sebastian has reported that this works better on his machine, so I'm going to merge.

PR #125, opened 2019-05-03
Check for config file in current directory

SebastianM-C:
With this change the current directory is also checked for a config file. This allows per-project config files.
The config file in the current directory takes precedence over the one in`config_path`.

### DIFF
--- a/src/AbstractPlotting.jl
+++ b/src/AbstractPlotting.jl
@@ -176,7 +176,8 @@ end
 
 const has_ffmpeg = Ref(false)
 
-const config_path = joinpath(homedir(), ".config", "makie", "theme.jl")
+const config_file = "theme.jl"
+const config_path = joinpath(homedir(), ".config", "makie", config_file)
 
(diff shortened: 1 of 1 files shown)


asinghvi17:
Changes don't seem to break anything, so merging.  Thanks, @SebastianM-C!

SebastianM-C:
I forgot to document this. I'll make a separate PR.

asinghvi17:
Where were you intending to document it?

SebastianM-C:
I think it should be mentioned here: http://juliaplots.org/MakieGallery.jl/stable/theming.html

asinghvi17:
Agreed, you'd need to see MakieGallery for that.  Have a look here: https://github.com/JuliaPlots/MakieGallery.jl/blob/master/docs/src/theming.md

SebastianM-C:
See https://github.com/JuliaPlots/MakieGallery.jl/pull/46

PR #126, opened 2019-05-04
fix config file parsing during precompilation (rollback #125)

asinghvi17:


asinghvi17:
From Slack via Sebastian:

>
> I realized that this is more complicated that I initially thought, The check for the local config file must be done at runtime, Due to precompilation the paths get "hardcoded", so if it finds one the local config file, it will always look for it. This will basically crash every package depending on `AbstractPlotting` that doesn't have a local config file.

PR #129, opened 2019-05-14
Don't constrain type of angle

goretkin:
This allows the use of UnitfulAngles

```julia
using AbstractPlotting
using UnitfulAngles
using StaticArrays

const τ = UnitfulAngles.turn

julia> AbstractPlotting.qrotation(@SVector[1,0,0], 0.25τ)
0.7071067811865476 + 0.7071067811865476im + 0.0jm + 0.0km

julia> AbstractPlotting.rotationmatrix_x(0)
4×4 SArray{Tuple{4,4},Float64,2,16}:
 1.0  0.0   0.0  0.0
 0.0  1.0  -0.0  0.0
 0.0  0.0   1.0  0.0
 0.0  0.0   0.0  1.0

julia> AbstractPlotting.rotationmatrix_x(0.5τ)
4×4 SArray{Tuple{4,4},Float64,2,16}:
 1.0   0.0   0.0  0.0
 0.0  -1.0  -0.0  0.0
 0.0   0.0  -1.0  0.0
 0.0   0.0   0.0  1.0
```

It doesn't seem like the `rotationmatrix_[]` methods are used in this package, but perhaps in a reverse dependency. So I wasn't sure if it was worth it to add a test.

### DIFF
--- a/src/camera/projection_math.jl
+++ b/src/camera/projection_math.jl
@@ -22,11 +22,11 @@ function translationmatrix(t::Vec{3, T}) where T
     )
 end
 
-rotate(angle::T, axis::Vec{3, T}) where {T} = rotationmatrix4(qrotation(convert(Array, axis), angle))
+rotate(angle, axis::Vec{3}) = rotationmatrix4(qrotation(convert(Array, axis), angle))
 rotate(::Type{T}, angle::Number, axis::Vec{3}) where {T} = rotate(T(angle), convert(Vec{3, T}, axis))
 
-function rotationmatrix_x(angle::T) where T
-    T0, T1 = zero(T), one(T)
+function rotationmatrix_x(angle)
+    T0, T1 = (0, 1)
     Mat{4}(
         T1, T0, T0, T0,
         T0, cos(angle), sin(angle), T0,
@@ -34,8 +34,8 @@ function rotationmatrix_x(angle::T) where T
         T0, T0, T0, T1
     )
 end
-function rotationmatrix_y(angle::T) where T
-    T0, T1 = zero(T), one(T)
+function rotationmatrix_y(angle)
+    T0, T1 = (0, 1)
     Mat{4}(
         cos(angle), T0, -sin(angle),  T0,
         T0, T1, T0, T0,
@@ -43,8 +43,8 @@ function rotationmatrix_y(angle::T) where T
         T0, T0, T0, T1
     )
 end
-function rotationmatrix_z(angle::T) where T
-    T0, T1 = zero(T), one(T)
+function rotationmatrix_z(angle)
+    T0, T1 = (0, 1)
     Mat{4}(
         cos(angle), sin(angle), T0, T0,
         -sin(angle), cos(angle),  T0, T0,
--- a/src/camera/projection_math.jl
+++ b/src/camera/projection_math.jl
@@ -25,7 +25,7 @@ end
 rotate(angle, axis::Vec{3}) = rotationmatrix4(qrotation(convert(Array, axis), angle))
 rotate(::Type{T}, angle::Number, axis::Vec{3}) where {T} = rotate(T(angle), convert(Vec{3, T}, axis))
 
-function rotationmatrix_x(angle)
+function rotationmatrix_x(angle::Number)
     T0, T1 = (0, 1)
     Mat{4}(
         T1, T0, T0, T0,
@@ -34,7 +34,7 @@ function rotationmatrix_x(angle)
         T0, T0, T0, T1
     )
 end
-function rotationmatrix_y(angle)
+function rotationmatrix_y(angle::Number)
     T0, T1 = (0, 1)
     Mat{4}(
         cos(angle), T0, -sin(angle),  T0,
@@ -43,7 +43,7 @@ function rotationmatrix_y(angle)
         T0, T0, T0, T1
     )
 end
-function rotationmatrix_z(angle)
+function rotationmatrix_z(angle::Number)
     T0, T1 = (0, 1)
     Mat{4}(
         cos(angle), sin(angle), T0, T0,
--- a/src/utilities/quaternions.jl
+++ b/src/utilities/quaternions.jl
@@ -34,7 +34,7 @@ end
 @inline Base.getindex(x::Quaternion, i::Integer) = x.data[i]
 Base.isapprox(x::Quaternion, y::Quaternion) = all((x.data .≈ y.data))
 
-function qrotation(axis::StaticVector{3}, theta::Real)
+function qrotation(axis::StaticVector{3}, theta)
     u = normalize(axis)
     s = sin(theta / 2)
     Quaternion(s * u[1], s * u[2], s * u[3], cos(theta / 2))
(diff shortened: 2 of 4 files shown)


src/camera/projection_math.jl
```diff
@@ -22,29 +22,29 @@ function translationmatrix(t::Vec{3, T}) where T
...
 end
 
-rotate(angle::T, axis::Vec{3, T}) where {T} = rotationmatrix4(qrotation(convert(Array, axis), angle))
+rotate(angle, axis::Vec{3}) = rotationmatrix4(qrotation(convert(Array, axis), angle))
 rotate(::Type{T}, angle::Number, axis::Vec{3}) where {T} = rotate(T(angle), convert(Vec{3, T}, axis))
 
-function rotationmatrix_x(angle::T) where T
-    T0, T1 = zero(T), one(T)
+function rotationmatrix_x(angle)
+    T0, T1 = (0, 1)
```
SimonDanisch: can we leave this? It's important to keep the type!

src/utilities/quaternions.jl
```diff
@@ -34,7 +34,7 @@ end
 @inline Base.getindex(x::Quaternion, i::Integer) = x.data[i]
 Base.isapprox(x::Quaternion, y::Quaternion) = all((x.data .≈ y.data))
 
-function qrotation(axis::StaticVector{3}, theta::Real)
+function qrotation(axis::StaticVector{3}, theta)
```
SimonDanisch: This seems to be the only real change needed here, right? Can we leave it at `Number`?

src/camera/projection_math.jl
goretkin: We certainly don't have to change it. But I am curious why not. The matrix will have a uniform type. The function as written will break, for example, for `rotationmatrix_x(1)`, since `sin(1)::Float64`. It will also break for `UnitfulAngles` numbers.

src/utilities/quaternions.jl
goretkin: That is probably fine. But what is the benefit of the `Number` type annotation? Either you can take the `sin`\`cos` of the type, or you can't, and you'll get an error too.

Don't know why you would want to here, but you can take the `sin` or square matrices, for example. I am generally in favor of unconstraining the types, so long as it does not hinder error reporting or dispatch. In this case, there's no dispatch needed, since there's only one definition of the `qrotation`method.

src/camera/projection_math.jl
asinghvi17: Don't you lose some generality by using 1 and 0 instead of one(T) and zero(T)?

src/utilities/quaternions.jl
SimonDanisch: after working with a lot of other people's code, I started following the rule to annotate types for documentation,.... Makes it so much easier to get started with a code base ;) And here, we even have the perfect type to annotate it with ;)

src/camera/projection_math.jl
goretkin: I don't think so, not in any important case. Though if someone can give a more satisfying answer, that would be better.


```
function all_subtypes(T)
  Ts= Set([T])
  while true
    n_before = length(Ts)
    more = Set(vcat(subtypes.(Ts)...))
    Ts = union(Ts, more)
    n_after = length(Ts)
    if n_after == n_before
      break
    end
  end
  return Ts
end

Numbers = collect(all_subtypes(Number))
Numbers = filter(!isabstracttype, Numbers)

[(old_type, promote_type(Int, old_type)) for old_type in Numbers if promote_type(Int, old_type) != old_type]
```

```
 (Bool, Int64)
 (UInt16, Int64)
 (Irrational, Float64)
 (Int8, Int64)
 (Unitful.Gain{L,S,T} where T<:Real where S where L<:Unitful.LogInfo, Number)
 (UInt32, Int64)
 (Unitful.Level{L,S,T} where T<:Number where S where L<:Unitful.LogInfo, Number)
 (UInt8, Int64)
 (Rational, Real)
 (Unitful.Quantity, Number)
 (Int16, Int64)
 (Complex, Number)
 (Int32, Int64)
```

SimonDanisch: Ah you're right.
This is the case I'm worried about:
```julia
angle = 1f0
T0, T1 = (0, 1)

x = Mat{4}(
           cos(angle), T0, -sin(angle),  T0,
           T0, T1, T0, T0,
           sin(angle), T0, cos(angle), T0,
           T0, T0, T0, T1
       )
@assert x isa Matf4f0
```
Which works fine with your changes :) I guess, Julia 1.0 changed the promotion rules!

src/utilities/quaternions.jl
goretkin: Perfect until someone invents a non-`Number` type that represents an angle! But that's unlikely.

I agree generally about type annotations helping as documentation, but not sure what else a user could think `theta` is in this case. But I think it does no harm to add `::Number`. I'll do that!

src/camera/projection_math.jl
goretkin: I can add this as a test, too, in case Julia ever changes it back! Where should it live? (Is this function used by `Makie`?)

src/utilities/quaternions.jl
SimonDanisch: Thank you :) It's a rule of thumb I like to follow in the code base wherever I can ... And if someone actually runs into a problem, it's very easy to remove the type constraint ;)

SimonDanisch:
Thanks a lot! :) If you can push the `::Number` change, this is good to go!!

src/camera/projection_math.jl
asinghvi17: That test would probably be better in `AbstractPlotting`'s tests, feel free to make a new file and testset :)

piever: It may be slightly cleaner to do `T0, T1 = (false, true)`  (which is the eltype of `I`) so that you are sure that the promotion doesn't increase the precision. Though in practice it would only matter if `cos(angle)` was for example `Int32` but that shouldn't happen.

goretkin: I think you mean `im` (`Complex(false, true)`)? I think you are technically right, though I prefer 0 and 1 aesthetically. I'll let someone else make the final call.

test/projection_math.jl
```diff
@@ -0,0 +1,2 @@
+@test eltype(AbstractPlotting.rotationmatrix_x(1)) == Float64
+@test eltype(AbstractPlotting.rotationmatrix_x(1f0)) == Float32
```
asinghvi17: Could you wrap this in a `testset`?  Something like:

```julia
@testset "Projection math" begin
    @test eltype(AbstractPlotting.rotationmatrix_x(1)) == Float64
    @test eltype(AbstractPlotting.rotationmatrix_x(1f0)) == Float32 
end

src/utilities/quaternions.jl
asinghvi17: Seems resolved by https://github.com/JuliaPlots/AbstractPlotting.jl/pull/129/commits/7d41fe03778ba2f246724d34a1dd0a2a547c8b0a

test/projection_math.jl
SimonDanisch: I changed it ;)

goretkin: Thanks! I didn't notice the quaternion tests in a test set (maybe I missed it?), so I wasn't sure what to do.

asinghvi17: They aren't yet, but I have a PR open for that

SimonDanisch:
Sorry to drag this simple PR out so much...
Thank you :)

PR #136, opened 2019-05-21
allow flipping

asinghvi17:


test/runtests.jl
```diff
@@ -1,9 +1,20 @@
 using AbstractPlotting
+using Makie
```
asinghvi17: Why `using Makie`?

SimonDanisch: Thanks, that shouldn't be there!

PR #137, opened 2019-05-23
 Fix save(path, io::VideoStream) when file type is .mkv 

tlqmj:
Previously, `save("video.mkv", io)` would fail because `out` wasn't defined.

I also changed the behavior when trying to overwrite an existing file. Previously, trying to overwrite a .mkv file would error out, while doing the same with any of the other supported formats would work just fine.

I'm new to this whole git business, so I was wondering if maybe I should have broken this second change into its own pull request?


asinghvi17:
Works on my machine, merging.  Thanks @tlqmj!

PR #138, opened 2019-05-23
Update `VideoStream`, `save` docstrings

tlqmj:
The docstring for `VideoStream`'s constructor had references to functions whose names and signatures had changed. 

It also had a lot of (outdated) information about how to use `save`. I deleted most of that as it was redundant with `save`'s docstring.


asinghvi17:
Thanks!

PR #140, opened 2019-05-26
Improve camera zooming

smldis:
Zoom is now done by focusing on the mouse pointer coordinates.
This improves greatly the usability of the user interface.

Other usability issues might be:
- Use fov instead of camera movement for zooming: this would avoid z-clipping problems at high zoom
- better defaults for rotationspeed and translation speed: I can test this only on my laptop and my settings migh be bad for others.
- rotation and translation speed seems to depend somehow on window size.

### DIFF
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -99,8 +99,6 @@ function add_translation!(scene, cam, key, button)
     end
     on(camera(scene), scene.events.scroll) do scroll
         if ispressed(scene, button[]) && is_mouseinside(scene)
-            # translate_cam!(scene, cam, Vec3f0(scroll[2], 0f0, 0f0))
-
             cam_res = Vec2f0(widths(scene.px_area[]))
             mouse_pos_normalized = Vec2f0(scene.events.mouseposition[]) ./ cam_res
             mouse_pos_normalized = 2*mouse_pos_normalized .- 1f0
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -155,6 +155,11 @@ function translate_cam!(scene::Scene, cam::Camera3D, _translation::VecTypes)
     return
 end
 
+"""
+    zoom!(scene, point, zoom_step)
+    
+Zooms the camera of `scene` in towards `point` by a factor of `zoom_step`.
+"""
 function zoomtopoint_cam!(scene, point, zoom_step)
     cam = cameracontrols(scene)
     @extractvalue cam (projectiontype, lookat, eyeposition, upvector)
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -160,7 +160,7 @@ end
     
 Zooms the camera of `scene` in towards `point` by a factor of `zoom_step`.
 """
-function zoomtopoint_cam!(scene, point, zoom_step)
+function zoom!(scene, point, zoom_step)
     cam = cameracontrols(scene)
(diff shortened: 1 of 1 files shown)


smldis:
Please consider this a draft, since I did not do a lot of testing, I don't know the Makie codebase and it's my first PR

smldis:
One issue: I don't understand why the zoom on 2D plots is reversed.

asinghvi17:
Thanks for the PR!

> Please consider this a draft

Just let us know when it's ready to merge ;)

> One issue: I don't understand why the zoom on 2D plots is reversed.

The zoom works on my machine for 2D plots, do you have unpushed commits?

Just a comment - may be nice to add a `zoom!` function to zoom in to a plot by some factor.

smldis:
@asinghvi17 This is the main issue now:
> The zoom works on my machine for 2D plots, do you have unpushed commits?
Yeah it works, but on my machine scrollwheel up is zooming out i.e. reversed for 2D plots (instead it is zooming in for the 3D plot as in every other 3D programs I use)

> Just a comment - may be nice to add a `zoom!` function to zoom in to a plot by some factor.
It is already in the PR, see `zoomtopoint_cam!(scene, point, zoom_step)`. Is there already any public API on camera movement now? If there is nothing I would consider this as a separate issue.

SimonDanisch:
This looks great!

>Is there already any public API on camera movement now?



Yeah, we should probably clean up what's already there, and make it an official camera API! ;)

src/camera/camera3d.jl
```diff
@@ -132,22 +139,41 @@ function translate_cam!(scene::Scene, cam::Camera3D, _translation::VecTypes)
...
     side_trans = right * (-x) + normalize(upvector) * y
-    newpos = eyeposition + side_trans + zoom_trans
+    newpos = eyeposition + side_trans + z_trans
     cam.eyeposition[] = newpos
     cam.lookat[] = lookat + side_trans
     update_cam!(scene, cam)
     return
 end
 
+function zoomtopoint_cam!(scene, point, zoom_step)
```
asinghvi17: Maybe change the name to `zoom!`, so that it is easier to access and interpret?

```suggestion
function zoom!(scene, point, zoom_step)
```

src/camera/camera3d.jl
```diff
@@ -132,22 +139,41 @@ function translate_cam!(scene::Scene, cam::Camera3D, _translation::VecTypes)
...
-    zoom_trans = dir_norm * zoom
+    z_trans = dir_norm * z
     side_trans = right * (-x) + normalize(upvector) * y
-    newpos = eyeposition + side_trans + zoom_trans
+    newpos = eyeposition + side_trans + z_trans
     cam.eyeposition[] = newpos
     cam.lookat[] = lookat + side_trans
     update_cam!(scene, cam)
     return
 end
```
asinghvi17:
```suggestion

"""
    zoom!(scene, point, zoom_step)
    
Zooms the camera of `scene` in towards `point` by a factor of `zoom_step`,
by reducing the field of view.
"""
```

SimonDanisch:
@smldis, should we merge this?

smldis:
@SimonDanisch I think this should be a blocker:
> One issue: I don't understand why the zoom on 2D plots is reversed.

Any idea? We can still merge and open an issue for later.

SimonDanisch:
>Any idea? We can still merge and open an issue for later.

That is not a regression of this PR right? Lets fix it later then

PR #142, opened 2019-06-08
[WIP] Refactor Axis3D

asinghvi17:
This PR aims to make it possible to theme an Axis3d similarly to an Axis2d, where the grid can have a linestyle.  This will make it easier to theme them and allow for 'less visible' grid lines in Themes.

This PR is related to JuliaPlots/MakieThemes.jl#10, and must be merged into master before its functionality can work.

TODO list (help wanted):
- [ ] Change API for Axis3D to be consistent with Axis2d (frames v/s grid)
- [ ] Determine which keyword attributes to include
- [ ] Add a tick formatter argument, so that you can have scientific ticks (first step to log scales?)
- [ ] Look into `PGFPlots`, `tufte` styling

### DIFF
--- a/src/basic_recipes/axis.jl
+++ b/src/basic_recipes/axis.jl
@@ -131,10 +131,16 @@ end
             font = lift(dim3, theme(scene, :font)),
         ),
 
+        grid = Theme(
+            linecolor = (grid_color, grid_color, grid_color),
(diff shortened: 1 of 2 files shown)


asinghvi17:
I'll postpone this till we merge MakieLayout, because I have some more ideas about it.  We need the ability to have axis frames, for example.

PR #144, opened 2019-06-10
make slider use raw and campixel by default

asinghvi17:


asinghvi17:
It feels a bit dirty, but could we set the Scene's `raw` and `camera` in the recipe itself?

SimonDanisch:
I merged & fixed this in `sd-new_version`

PR #151, opened 2019-06-24
Better colormap resolution

asinghvi17:


asinghvi17:
This is what `cmocean` looks like with this layout.
cmocean

PR #152, opened 2019-06-25
Test on Gitlab and implement "OpenGL-aware" testing

asinghvi17:


asinghvi17:
Will also need to integrate into test/runtests.jl.

asinghvi17:
Now, the only remaining question is - what's a good default to have, minimal or full tests?

asinghvi17:
This particular commit should fail CI, because the reference images need to be updated.  This is normal and expected.

asinghvi17:
@SimonDanisch, would it make sense to have Gitlab open a PR to SImonDanisch/ReferenceImages with updated reference images if a PR on AbstractPlotting gets merged?  It could keep pushing to a single branch (maybe `gitlab/update-refimages` or something).

asinghvi17:
I think this works as is, so I'll rebase the branch and merge.

PR #157, opened 2019-07-06
Test cam3d! changes

asinghvi17:
cc @CarpeNecopinum


asinghvi17:
The only example that this breaks is the lorenz attractor one, which is just because the way the mouse input is interpreted has changed.

Issue #147, opened 2019-06-18
Creating CAD style camera handling


Kevin-Mattheus-Moerman:
This relates to: 
https://github.com/JuliaGL/GLAbstraction.jl/issues/96 (and also: https://github.com/JuliaPlots/Makie.jl/issues/38, https://github.com/JuliaGL/GLAbstraction.jl/issues/76, https://github.com/JuliaGL/GLAbstraction.jl/pull/80, https://github.com/JuliaPlots/Makie.jl/issues/329). 

Moving discussion here as it is about AbstractPlotting's camera handling (in particular https://github.com/JuliaPlots/AbstractPlotting.jl/blob/master/src/camera/camera3d.jl)

Kevin-Mattheus-Moerman:
The goals is to have AbstractPlottings camera handling be like this (screenshot of MeshLab): 

![cat](https://user-images.githubusercontent.com/8392709/59666588-058ce400-91ad-11e9-8ce1-658ede0a7b3f.gif)
 i.e. such that moving the cursor to the left or right (screen's/viewer's [1 0] direction) rotates the object around the screen's/viewer's vertical axis ([0 1]). Similarly moving the cursor up or down (screen's/viewer's [0 1] direction) will rotate the object around the screen's/viewer's left-right or horizontal axis ([1 0]).

CarpeNecopinum:
Replacing
```rotation = rotate_cam(theta_v, right_v, Vec3f0(0, 0, sign(upvector[3])), dir)```
in line 189 by
```rotation = rotate_cam(theta_v, right_v, upvector, dir)```
already gets us most of the way there.

Additionally I suggest removing line 174 (`cam.lookat[] = lookat + zoom_translation`) for a more CAD-style feeling, as it allows rotation of a zoomed-in object.

If desired, I can open a PR for this.

CarpeNecopinum:
In general I think we should allow multiple different 3d camera variants to exist in AbstractPlotting.jl, as the old behavior is probably better for "just a plot"-style scenes.

On the other end of the spectrum, I prefer a FPS/Spaceship-style camera movement (like in https://github.com/CarpeNecopinum/Shiny.jl/blob/master/src/movement.jl) for large, complex scenes.

asinghvi17:
@CarpeNecopinum, that would be great!

Kevin-Mattheus-Moerman:
:clap: @CarpeNecopinum please do

[...]

PR #161, opened 2019-07-11
Added cam3d_cad!

CarpeNecopinum:
Implemented the suggestion of two 3d camera functions from #147.

### DIFF
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -16,8 +16,7 @@ struct Camera3D <: AbstractCamera
     move_key::Node{ButtonTypes}
 end
 
-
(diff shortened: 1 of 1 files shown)


asinghvi17:
Let's wait to merge this into master, until the new version stuff is settled?  Changes look good though!

CarpeNecopinum:
new version stuff? 

As suggested, I left cam_3d! as alias to cam_3d_turntable!, so this change shouldn't break existing code.

asinghvi17:
True, I'll check in some time whether there are any merge conflicts, but I don't think any of the camera code has been touched.  If that's the case, then I think this should be good to merge in master, if people agree with the changes.

PR #172, opened 2019-07-23
Scale colorlegend to the scene resolution

asinghvi17:


asinghvi17:
This is what the colormap looks like with this PR:
lol

src/basic_recipes/legend.jl
```diff
@@ -34,10 +36,10 @@ creates a legend from a colormap
 """
 @recipe(ColorLegend, colormap, colorrange) do scene
     Theme(
-        width = (20, 200),
+        width = (20, lift(x -> x[2], theme(scene, :resolution))),
```
SimonDanisch: should rather use `lift(x-> width(x)[2], pixelarea(scene))`.. The theme value won't update when resizing the window

asinghvi17:
```suggestion
        width = (20, lift(x -> x[2], pixelarea(scene))),

File: src/basic_recipes/legend.jl [hunk 2]
```diff
@@ -34,10 +36,10 @@ creates a legend from a colormap
 """
 @recipe(ColorLegend, colormap, colorrange) do scene
     Theme(
-        width = (20, 200),
+        width = (20, lift(x -> x[2], pixelarea(scene))),
```
asinghvi17:
```suggestion
        width = (20, lift(x -> x.h, pixelarea(scene))),
```

PR #186, opened 2019-07-31
Make legends work with multiplots

asinghvi17:
Flatten the plot structure to remove all multiplots!

### DIFF
--- a/src/basic_recipes/basic_recipes.jl
+++ b/src/basic_recipes/basic_recipes.jl
@@ -369,7 +369,7 @@ end
 
 
 """
-    `annotations(strings::Vector{String}, positions::Vector{Point})`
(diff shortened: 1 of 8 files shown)


SimonDanisch:
What's the state of this? :)

asinghvi17:
Not really working...the idea was to flatten the plot structure, but I think I'll wait until layouting is integrated - it seems pretty close now!  After that we can decide if we should restructure the plot struct a little, to account for labels.  We could also change the legend's structure to allow for nested labels.

asinghvi17:
Well it was fun, but I think a better idea is to look for a "label" attribute - this will allow us to support nested structures as well.  Getting good legend drawing will be quite elaborate, if we want to support all kinds of things + "interactive" / manual legend creation.

PR #191, opened 2019-08-15
Tickmarks

grero:
This adds some code to add tickmarks to a 2D axis. Improvements and suggestions are of course most welcome. This PR basically adds the `draw_tickmarks` function, in which I combined code from `draw_tick`s and `draw_grid`. I also added a section `tickmarks` to the default Axis2D theme, as well as a `showtickmarks` property.

### DIFF
--- a/src/AbstractPlotting.jl
+++ b/src/AbstractPlotting.jl
@@ -129,6 +129,7 @@ export textslider
 
 # gui
 export slider, button, playbutton
+export move!
 
 # Raymarching algorithms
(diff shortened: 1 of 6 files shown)


grero:
I did a git rebase -I here to reorder some of my commits and it seems I also touched some previous commits that are obviously not mine. Does anyone know how I can fix it so that those are not part of this PR?

asinghvi17:
@grero, you could probably git-cherrypick onto a new branch from master, seems like the easiest way.

PR #192, opened 2019-08-16
Tickmarks2

grero:
Cleaned-up version of #191.


SimonDanisch:
Is this good to merge?

grero:
I have been using it for my own work and it seems to do the job. Note that it doesn't include tick marks for 3D plots. I could add those here, or I can open a separate PR for that. It's just that I haven't been using 3D plots much lately and so was focused on 2D plots only.

PR #195, opened 2019-08-29
fix limit for 2d

SimonDanisch:


asinghvi17:
Should we register a 0.9.10 with this...or just wait for 1.0?  Passing resolution tuple to the Scene is pretty common...

PR #211, opened 2019-10-28
Improve enum and v/hbox docs

asinghvi17:
Fixes JuliaPlots/MakieGallery.jl#155

### DIFF
--- a/src/documentation/docstringextension.jl
+++ b/src/documentation/docstringextension.jl
@@ -2,6 +2,10 @@
 #                              DocStringExtension                              #
 ################################################################################
 
+############################################################
+#                        Attributes                        #
+############################################################
+
 struct DocThemer <: DocStringExtensions.Abbreviation end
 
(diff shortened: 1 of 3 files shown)


logankilpatrick:
Something I just noticed today: When `vbox`ing two scenes together with different limits, it defaults to make the `vbox` the size of the largest one. This can lead to an issue similar to what I was having in https://github.com/JuliaPlots/Makie.jl/issues/397 

I would add a note like: 

"when using `vbox` to attach two scenes together, it is recommmented that the scenes have the same scale/size/limits(not sure of the correct word here) in order to make sure that the two scenes in the vbox/hbox are proportional to one another."

SimonDanisch:
This is ready to merge, is it?

asinghvi17:
IDK why Travis fails, but should be.  Also there's a blackout in my area, so I may be out of touch for a while - feel free to merge this.

@logankilpatrick, I'll incorporate that change next PR I make.

PR #219, opened 2019-11-25
add textspace and right align label so sliderwidth includes the label

jkrumbiegel:
this helps putting the slider in a layout. the textspace must be set manually so that it makes sense for the values used. a convenience function could be made that adjusts for the current label size once.


jkrumbiegel:
grafik

this is how the slider looks in a layout with the fix

SimonDanisch:
Looks good :)

PR #247, opened 2019-12-08
add @lift to reduce redundancy in lift() calls

jkrumbiegel:
Replaces an expression with lift(argtuple -> expression, args...), where args
are all expressions inside the main one that begin with $.

```julia
# Example:

x = Node(rand(100))
y = Node(rand(100))

## before
z = lift((x, y) -> x .+ y, x, y)

## after
z = @lift($x .+ $y)
```

### DIFF
--- a/src/AbstractPlotting.jl
+++ b/src/AbstractPlotting.jl
@@ -31,6 +31,7 @@ include("utilities/utilities.jl")
 include("utilities/logging.jl")
 include("utilities/texture_atlas.jl")
 include("interaction/nodes.jl")
+include("interaction/liftmacro.jl")
 
 # Basic scene/plot/recipe interfaces + types
 include("scenes.jl")
@@ -85,7 +86,7 @@ export xlims!, ylims!, zlims!
 export xlabel!, ylabel!, zlabel!
 
 # Node/Signal related
-export Node, node, lift, map_once, to_value, on
+export Node, node, lift, map_once, to_value, on, @lift
 
 # utilities and macros
 export @recipe, @extract, @extractvalue, @key_str, @get_attribute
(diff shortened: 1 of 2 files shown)


asinghvi17:
Observables already has a macro that does something similar - 
```julia
  @map(expr)

  Wrap AbstractObservables in & to compute expression expr using their value. The expression will be computed when @map is called and every time the AbstractObservables are updated.

  Examples
  ==========

  julia> a = Observable(2);

  julia> b = Observable(3);

  julia> c = Observables.@map &a + &b;

  julia> c[]
  5

  julia> a[] = 100
  100

  julia> c[]
  103
```
Perhaps we should alias that?  Or we could keep both, and document them.

jkrumbiegel:
ah interesting, that does seem to do exactly the same on first glance. I only looked if there was something comparable in AbstractPlotting

jkrumbiegel:
ah actually they seem to work differently, check this out
grafik

aside from the difference in macro hygiene (mine should be functionally the same even though there is variable shadowing in the closure) the getindex call is not transfered into the closure, but stays with the argument, which would error for an observable. So maybe my macro is better for observables

asinghvi17:
Interesting, I didn't notice that.  Is this good to merge?

jkrumbiegel:
as far as I can see it does what it's supposed to and I've documented it a bit as well, but I'm not super experienced with macros so I don't know if there might be any gotchas

asinghvi17:
@jkrumbiegel could you document this in MakieGallery/docs/src/interaction.md?

jkrumbiegel:
done! https://github.com/JuliaPlots/MakieGallery.jl/pull/167

asinghvi17:
Thanks!

PR #251, opened 2019-12-10
implement delete plot from scene

SimonDanisch:


SimonDanisch:
Needs: https://github.com/JuliaPlots/GLMakie.jl/pull/54

PR #253, opened 2019-12-11
Remove Any type in convert definition to Node

asinghvi17:


SimonDanisch:
Nodes got reworked in #315

PR #270, opened 2020-01-13
Export units

asinghvi17:
@SimonDanisch could you comment on which units work?  I haven't had any luck with rel and mm as of now.


SimonDanisch:
Yeah only px should work... And currently just with master of GLMakie

PR #283, opened 2020-01-26
Convenience functions for manipulating ticks

Tokazama:
This covers tick rotation and labels. One of the goals of this PR is to create an interface that should be consistent no matter what changes might happen to the internals of the `Scene` type. Theoretically the only changes that would have to be made to make all of this code work in the presence of other backend changes would be to the `ticks!` and `ticksrotations!`.

It's is still a WIP. I plan on creating tests for all the code I'm adding once everyone is satisfied with method name and other syntax.

### DIFF
--- a/src/shorthands.jl
+++ b/src/shorthands.jl
@@ -6,7 +6,7 @@ Defaults to using the current Scene.
 """
 function xlabel!(scene, xlabel::AbstractString)
     axis = scene[Axis]
-    @assert !isnothing(axis)
+    @assert !isnothing(axis) || "The Scene does not have an axis!"
     scene[Axis][:names][:axisnames][] = (xlabel, scene[Axis][:names][:axisnames][][2:end]...)
     nothing
 end
@@ -20,7 +20,7 @@ Defaults to using the current Scene.
 """
 function ylabel!(scene, ylabel::AbstractString)
     axis = scene[Axis]
-    @assert !isnothing(axis)
+    @assert !isnothing(axis) || "The Scene does not have an axis!"
     if axis isa Axis2D
         scene[Axis][:names][:axisnames][] = (scene[Axis][:names][:axisnames][][1], ylabel)
     elseif axis isa Axis3D
@@ -42,8 +42,8 @@ Defaults to using the current Scene.
 """
 function zlabel!(scene, zlabel::AbstractString)
     axis = scene[Axis]
-    @assert !isnothing(axis)
-    @assert axis isa Axis3D
+    @assert !isnothing(axis) || "The Scene does not have an axis!"
+    @assert axis isa Axis3D || "The scene does not have a z-axis"
     scene[Axis][:names][:axisnames][] = (scene[Axis][:names][:axisnames][][1], scene[Axis][:names][:axisnames][][2], zlabel)
     nothing
 end
@@ -112,7 +112,7 @@ Returns the all the axis tick labels.
 """
 function ticklabels(scene)
     axis = scene[Axis]
-    @assert !isnothing(axis)
+    @assert !isnothing(axis) || "The Scene does not have an axis!"
     return axis.ticks.ranges_labels[][2]
 end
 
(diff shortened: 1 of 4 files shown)


src/shorthands.jl
```diff
@@ -104,3 +104,215 @@ zlims!(lims::NTuple{2, Real}) = zlims!(current_scene(), lims)
...
+function ticks!(; tickranges=tickranges(scene), ticklabels=ticklabels(scene))
+    ticks!(current_scene(), ranges=ranges, labels=labels)
+end
+
+"""
+    xticks!([scene,]; xtickranges=xtickrange(scene), xticklabels=xticklabel(scene))
+
+Set the tick labels and range along the x-axes. See also [`ticks!`](@ref).
+"""
+function xticks!(scene; xtickrange=xtickrange(scene), xticklabels=xticklabels(scene))
```
asinghvi17:
```suggestion
function xticks!(scene = current_scene(); xtickrange=xtickrange(scene), xticklabels=xticklabels(scene))
```
so on and so forth...

src/shorthands.jl
```diff
@@ -104,3 +104,215 @@ zlims!(lims::NTuple{2, Real}) = zlims!(current_scene(), lims)
...
+
+###
+### Ticks rotations
+###
+"""
+    ticksrotations(scene)
+
+Returns the rotation of all tick labels.
+"""
+function ticksrotations(scene)
```
asinghvi17: Maybe this should be renamed to `tickrotations`?

src/shorthands.jl
```diff
@@ -104,3 +104,215 @@ zlims!(lims::NTuple{2, Real}) = zlims!(current_scene(), lims)
...
+    axis = scene[Axis]
+    @assert !isnothing(axis)
```
asinghvi17:
```suggestion
    @assert !isnothing(axis) "The Scene does not have an axis!"
```

src/shorthands.jl
```diff
@@ -104,3 +104,215 @@ zlims!(lims::NTuple{2, Real}) = zlims!(current_scene(), lims)
...
+Returns the rotation of tick labels along the z-axis. See also [`ticksrotations`](@ref)
+"""
+zticksrotation(scene) = ticksrotations(scene)[3]
+
+"""
+    xticksrotation!([scene,] zangle)
+
+Set the rotation of all tick labels.
+"""
+function ticksrotations!(scene, angles)
```
asinghvi17: Maybe rename to `tickrotations!`?

src/shorthands.jl
```diff
@@ -104,3 +104,215 @@ zlims!(lims::NTuple{2, Real}) = zlims!(current_scene(), lims)
...
+    scene[Axis].ticks.rotation[] = angles
+    return nothing
+end
+
+"""
+    xticksrotation!([scene,] xangle)
+
+Set the rotation of tick labels along the x-axis. See also [`ticksrotations!`](@ref).
+"""
+function xticksrotation!(scene, xangle)
```
asinghvi17:
```suggestion
function xtickrotation!(scene=current_scene(), xangle)

Comment (Tokazama) [reply src/shorthands.jl]: This one will only work if `xangle` becomes a keyword argument. Should I change it to a keyword arg.

File: src/shorthands.jl (hunk 3 above)
Comment (Tokazama) [reply src/shorthands.jl]: Would you like me to add this to all of the functions in this file, not jus the methods I'm contributing.

File: src/shorthands.jl (hunk 5 above)
Comment (asinghvi17) [reply src/shorthands.jl]: Ah no, nevermind then - makes sense.  We should probably restrict `scene::Scene` then, as well, for the purposes of dispatch.

File: src/shorthands.jl (hunk 3 above)
Comment (asinghvi17) [reply src/shorthands.jl]: May as well...thanks!

File: src/shorthands.jl (hunk 1 above)
Comment (Tokazama) [reply src/shorthands.jl]: Are keyword arguments alright here? It was the simplest solution I had for managing setting the need to set both the range and labels at the same time.

Comment (asinghvi17) [reply src/shorthands.jl]: Yeah, kwargs should be fine.  Note that there's a new tick system in MakieLayout which sort of changes the internals of the function (though not the method signature).

Comment (Tokazama) [reply src/shorthands.jl]: I'm assuming that the need to change both will remain in the new system?

Comment (asinghvi17) [reply src/shorthands.jl]: Yeah, you still will be able to change both, but once you do, zooming won't work properly until you reset the ticks to the default.

File: test/shorthands_tests.jl [hunk 6]
```diff
@@ -0,0 +1,37 @@
...
+    @test xtickrotation(scene) == 0.0
+
+    ytickrotation!(0.0)
+    @test ytickrotation(scene) == 0.0
+
+    ztickrotation!(0.0)
+    @test ztickrotation(scene) == 0.0
+end
+
+
```
asinghvi17: Could you add some `@test_throws` here as well for trying to set a 2D Scene's z-axis attributes?

Tokazama:
Tests should be in place at this point. Any other changes necessary for this?

asinghvi17:
Would be nice to have a complementary PR on MakieGallery about the available convenience functions.  I can also contribute to that when I have time.

Tokazama:
I'll create an issue that maps out what that could be like. I'm sure it will be straight forward but I have other obligations this week.

PR #307, opened 2020-02-14
loosen constraints of `to_colormap`

### LINKED ISSUE #306: allow reversing colormaps and other colormaps with MakieLayout
I want to use the reverse colormap of e.g. `:deep`. Makie suggests that this is done by calling `Reverse(:deep)`. However makielayout has problems with that:
```
ERROR: LoadError: MethodError: no method matching to_colormap(::Reverse{Symbol}, ::Int64)
Closest candidates are:
  to_colormap(::Union{String, Symbol}, ::Integer) at C:\Users\m300808\.julia\packages\AbstractPlotting\wnYjV\src\conversions.jl:693
  to_colormap(::Any) at C:\Users\m300808\.julia\packages\AbstractPlotting\wnYjV\src\conversions.jl:23
Stacktrace:
 [1] (::MakieLayout.var"#316#324"{Int64})(::Reverse{Symbol}) at C:\Users\m300808\.julia\packages\MakieLayout\3hR9w\src\lobjects\lcolorbar.jl:65
 [2] lift(::Function, ::Observables.Observable{Any}) at C:\Users\m300808\.julia\packages\AbstractPlotting\wnYjV\src\interaction\nodes.jl:9
```

(the above is from a call to surface).

Custom colormaps, e.g. `cmap = to_colormap(:viridis, 100)` also fail with such surface plots _and_ MakieLayout.

asinghvi17:
Fix #306


asinghvi17:
I'll wait for Gitlab, then merge.

src/conversions.jl
```diff
@@ -667,12 +667,12 @@ struct Reverse{T}
     data::T
 end
 
-function convert_attribute(r::Reverse, ::key"colormap")
-    reverse(to_colormap(r.data))
+function convert_attribute(r::Reverse, ::key"colormap", n::Integer = 20)
```
SimonDanisch: Is this also needed? This can't really work I think....

asinghvi17: why not?

SimonDanisch: Huh, maybe it's the right way :D

asinghvi17:
@SimonDanisch, any idea why the gitlab runner gets an MbedTLS error?  Is there a way we can purge it?

asinghvi17:
Whoo, Gitlab is green!

asinghvi17:
It looks like there's a package resolution error on 1.3, potentially caused by StatsMakie.  @piever would adding Tables 1.0 resolve this?

PR #310, opened 2020-02-16
Fix conversion for a function on an interval

piever:


asinghvi17:
Needs a MakieGallery test (possibly also one in AbstractPlotting)

piever:
I'm not too sure how MakieGallery works, but I can definitely add tests for the conversion here in AbstractPlotting.

asinghvi17:
I've added some small tests in JuliaPlots/MakieGallery.jl#213

asinghvi17:
Will wait for Gitlab, then merge.  That should also be the first test of dynamic refimage update.

piever:
Thanks! Could you also tag a patch release afterwards? That way I can tag StatsMakie

PR #315, opened 2020-02-19
remove all the Observable conversion clutter

SimonDanisch:
Needs: https://github.com/JuliaGizmos/Observables.jl/pull/40


asinghvi17:
I don't think Observables reexports the new stuff yet.

SimonDanisch:
It shouldn't need to, I think^^
Did you remove most usage of `push!` from makiegallery?

asinghvi17:
Yeah, the only uses of it were in the tutorials anyway...we may not import the functions from Observables explicitly, though.

Issue #317, opened 2020-02-21
Issue with tick setters and automatic limits


asinghvi17:
They don't work, since the limits need to be materialized.  Not sure what's to be done here...

SimonDanisch:
Can you give an example of what's happening?

asinghvi17:
```julia
scene = Scene()
scene.limits[]
# AbstractPlotting.Automatic()
```

Then, when the `setlims!` function tries to get the current limits, it ends up getting `automatic` instead.

PR #319, opened 2020-02-22
fix setlims

SimonDanisch:
fix: https://github.com/JuliaPlots/AbstractPlotting.jl/issues/317


SimonDanisch:
>Isovalue 

That should be fixed in abstractplotting, right?

asinghvi17:
I thought that fix didn't work, but I guess I was wrong...?  Anyway, I'll merge and we'll see what the CI on master looks like.  Thanks!

PR #320, opened 2020-02-23
CI for text rewrite

asinghvi17:


asinghvi17:
Damn it...gitlab is working but Travis is now failing with that same package resolution error (even though I'm not using `--project`)

PR #350, opened 2020-03-06
Make some fixes to record and friends

asinghvi17:
- Provides a way to disable the auto-sleep when recording
cc @jessebett

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -407,8 +407,8 @@ end
 
 
 """
-    record(func, scene, path; framerate = 24, kwargs...)
-    record(func, scene, path, iter; framerate = 24, sleep = true, kwargs...)
+    record(func, scene, path; framerate = 24)
+    record(func, scene, path, iter; framerate = 24, sleep = true)
 
 Records the Scene `scene` after the application of `func` on it for each element
 in `itr` (any iterator).  `func` must accept an element of `itr`.
@@ -422,6 +422,11 @@ extension.  Allowable extensions are:
 
 `.mp4` and `.mk4` are marginally bigger and `.gif`s are up to
 6 times bigger with the same quality!
+
+The `sleep` keyword argument controls the rate at which frames are generated.
+When false, frames are generated and recorded as fast as the backend can generate
+them; when true, there is an artificial wait such that the animation plays on the
+screen in real time as it is being recorded.
 
 Typical usage patterns would look like:
 
@@ -451,9 +456,7 @@ function record(func, scene, path; framerate::Int = 24, kwargs...)
     save(path, io; framerate = framerate, kwargs...)
 end
 
-const _SLEEP_WHEN_RECORDING = Ref(true)
-
-function record(func, scene, path, iter; framerate::Int = 24, sleep = _SLEEP_WHEN_RECORDING[], kwargs...)
+function record(func, scene, path, iter; framerate::Int = 24, sleep = get(ENV, "MAKIE_SLEEP_WHEN_RECORDING", "true") == "true")
     io = VideoStream(scene; framerate = framerate)
     for i in iter
         t1 = time()
--- a/src/display.jl
+++ b/src/display.jl
@@ -399,7 +399,7 @@ function save(path::String, io::VideoStream;
         rm(pname, force = true)
     else
         rm(io.path)
-        @error("Video type $typ not known")
+        error("Video type $typ not known")
     end
     rm(io.path)
     return path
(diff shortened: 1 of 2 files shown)


asinghvi17:
@SimonDanisch do these changes make sense, at least for mp4s?

src/display.jl
```diff
@@ -386,16 +393,16 @@ function save(path::String, io::VideoStream;
         rm(pname, force = true)
     else
         rm(io.path)
-        error("Video type $typ not known")
+        @error("Video type $typ not known")
```
asinghvi17, line 396:
```suggestion
        error("Video type $typ not known")

**Comment by asinghvi17** [comment]
@SimonDanisch mind taking a look at this?  Changes are pretty minimal...the bitrate stuff didn't work out, and I still have to look into that, but this should speed up our CI times significantly :D

File: src/display.jl [hunk 2]
```diff
@@ -437,55 +444,26 @@ end
...
 
-```example
-    scene = lines(rand(10))
-    record(scene, "test.gif", 1:100) do i
-        scene.plots[:color] = Colors.RGB(i/255, 0, 0) # animate scene
-    end
-```
-"""
-function record(func, scene, path, iter; framerate::Int = 24)
+function record(func, scene, path, iter; framerate::Int = 24, sleep = _SLEEP_WHEN_RECORDING[], kwargs...)
```
SimonDanisch, line 456: Whats this `_SLEEP_WHEN_RECORDING` thing?

src/display.jl
```diff
@@ -400,8 +407,8 @@ end
 
 
 """
-    record(func, scene, path; framerate = 24)
-    record(func, scene, path, iter; framerate = 24)
+    record(func, scene, path; framerate = 24, kwargs...)
```
SimonDanisch, line 410: What are these kwargs for? They should be written out or documented!

src/display.jl
asinghvi17: I couldn't think of a better way to allow the default value of `sleep` to be changed, so I made it a Ref.  Thinking about it now, we should probably use an env variable instead...

src/display.jl
asinghvi17: Will do, I left that in from when I had 5 bitrate-related kws :D

src/display.jl
```diff
@@ -355,7 +355,10 @@ function recordframe!(io::VideoStream)
 end
 
 """
-    save(path::String, io::VideoStream; framerate = 24)
+    save(
```
SimonDanisch, line 358: Can we revert this style change? I prefer the one used before ;)

SimonDanisch: I should really add a style guide for Makie & friends :(

src/display.jl
SimonDanisch: What's the use case to change it globally?

asinghvi17: MakieGallery mainly - we want users to see the animation if they run it, but we also want to record it quickly on CI.

src/display.jl
asinghvi17: Would definitely help...I think I did this because there were too many kwargs, and didn't revert when all was done.

src/display.jl
SimonDanisch: If we want a mode, that records as fast as possible without adhering to a framerate, we should add that to the API and use that in `record` in our examples!

asinghvi17: Should we make this the default?

SimonDanisch: Sorry to be so naggy, but messing up our API with confusing hacks like this is simply not worth the short term gain ;) How is any outside collaborator able to interpret this ENV variable when trying to update the code? How would anyone figure out how to replace this ENV in a refactor that tries to make the API more consistent? 
We really need to design things with maintaining & documenting our API as the main bottleneck in mind, instead of sacrificing the former for improving our CI, or getting a quick hack in to allow high dpi support^^ 
I'm talking from experience here, since I'm the biggest offender of this principle, and everytime I did something like this, I got really sad 1 year later when trying to refactor the CODE and clean up all the hacks.

asinghvi17: Fair enough :) - this solution is pretty hacky no matter which way you slice it.

asinghvi17: Also, it doesn't look like this impacts CI time too much, so we could probably skip updating the Gallery...

SimonDanisch:
Can this be closed in favor of https://github.com/JuliaPlots/AbstractPlotting.jl/pull/383?

asinghvi17:
Yes, I think so.

PR #356, opened 2020-03-13
Squash error if .julia exists

asinghvi17:


src/conversions.jl
```diff
@@ -479,6 +479,55 @@ function convert_arguments(
...
+    x, y = PlotUtils.adapted_grid(f, endpoints(i))
+    ptype = plottype(P, Lines)
+    to_plotspec(ptype, convert_arguments(ptype, x, y))
+end
+
+to_tuple(t::Tuple) = t
+to_tuple(t) = (t,)
+
+function convert_arguments(P::PlotFunc, f::Function, args...; kwargs...)
+    tmp = f(args...; kwargs...) |> to_tuple
```
SimonDanisch: to_tuple should already exist somewhere, and please dont use `|>`

asinghvi17: That's the only place it's defined, when I removed it and forgot to paste back I got an undefvar error for it.
```suggestion 
    tmp =to_tuple(f(args...; kwargs...))

asinghvi17: We could relocate it somewhere else?

SimonDanisch: Ah lol, you just moved it around - i guess I meant exactly that definition then :D well, then lets just remove the pipe :D

asinghvi17: done :)

SimonDanisch:
Did you try running the code that uses `tryrange` ?

asinghvi17:
Haven't used it in anything yet, I think that got in by accident

PR #358, opened 2020-03-13
Remove tryrange

asinghvi17:
There's a blocking dispatch which is needed for StatsMakie, so `plot(sin)` can't work.  What I could do is `plot(automatic, sin)`, but that doesn't seem like a good API...

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -500,29 +500,6 @@ function convert_arguments(P::PlotFunc, f::Function, args...; kwargs...)
     convert_arguments(P, tmp...)
 end
 
-# The following `tryrange` code was copied from Plots.jl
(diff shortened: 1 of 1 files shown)


asinghvi17:
This is the implementation I was going to use:
```julia
function convert_arguments(p::PointBased, ::Automatic, f::Function)
    xmin = tryrange(f, [-5,-1,0,0.01]) # establish a minimum where the function doesn't error
    xmax = tryrange(f, filter(x -> x > xmin, [5,1,0.99, 0, -0.01])) # same for a maximum

    return PlotUtils.adapted_grid(f, (xmin, xmax) # this returns (x, f.(x)) which goes through another convert
end
```

SimonDanisch:
That looks like a good idea :)

asinghvi17:
Will make another PR to implement that, then!

SimonDanisch:
Thank you! :)

PR #360, opened 2020-03-13
document & implement refactor goals

SimonDanisch:


src/convert_arguments.jl
```diff
@@ -0,0 +1,996 @@
...
+
+# ColorBrewer colormaps that support only 8 colors require special handling on the backend, so we show them here.
+const colorbrewer_8color_names = String.([
+    :Accent,
+    :Dark2,
+    :Pastel2,
+    :Set2
+])
+
+# throw an error i
```
asinghvi17:
```suggestion

File: src/convert_attributes.jl [hunk 2]
```diff
@@ -0,0 +1,427 @@
...
+            return to_colormap(ColorBrewer.palette(cs_string, 8), n)
+        else                                    # cs_string must be in plotutils_names
+            return RGBf0.(PlotUtils.cvec(Symbol(cs), n))
+        end
+    else
+        error("There is no color gradient named: $cs")
+    end
+end
+
+function AbstractPlotting.convert_attribute(cg::PlotUtils.ColorGradient, ::key"colormap", n::Integer=length(cg.values))
```
asinghvi17: We should move this up near the rest of the color code.

src/AbstractPlotting.jl
```diff
@@ -1,6 +1,6 @@
 module AbstractPlotting
 
-using FFMPEG # get FFMPEG on any system!
++using FFMPEG # get FFMPEG on any system!
```
asinghvi17:
```suggestion
using FFMPEG # get FFMPEG on any system!

File: src/units.jl [hunk 4]
```diff
@@ -1,3 +1,23 @@
...
+## Move attributes into certain unit spaces
+# Future:
+# Every plot / scene axis will hold a space attribute
+# with the following allowed (per axis):
+# (may not actually be a type hierarchy, just here for illustration)
+abstract type DataSpace end
+abstract type DisplaySpace end
+#...
+struct Log <: DataSpace end
+struct Polar <: DataSpace end
```
asinghvi17: Most nonlinear projections will need their own axis types...this is something to discuss.

src/scenes.jl
```diff
@@ -1,14 +1,19 @@
 
 """
-    Scene TODO document this
+    A Scene can be understood as a (sub) window.
+    It opens up a rectangle in pixels (pixelarea(scene)) on the current screen,
+    and a camera that maps from data space to the pixel in the rectangle.
+    It also holds all graphical objects that are drawn in that area.
+    It currently also holds the display events (mouse position etc),
+    but this may change!
```
asinghvi17:
```suggestion
    Scene
    
A Scene can be understood as a (sub) window.
It opens up a rectangle in pixels (`pixelarea(scene)`) on the current screen,
and a camera that maps from data space to the pixels in the rectangle.
It also holds all graphical objects that are drawn in that area.
Currently, it holds the display events (mouse position, etc.),
but this may change!
```

src/scenes.jl
```diff
@@ -1,14 +1,19 @@
...
 
 ## Constructors
 $(SIGNATURES)
 
 ## Fields
 $(FIELDS)
 """
-mutable struct Scene <: AbstractScene
+struct struct Scene <: AbstractScene
```
asinghvi17:
```suggestion
struct Scene <: AbstractScene
```

src/theming.jl
```diff
@@ -45,9 +45,11 @@ const minimal_default = Attributes(
     camera = automatic,
     limits = automatic,
     padding = Vec3f0(0.1),
-    raw = false
+    raw = false,
 )
 
+plot_name(::AbstractPlot)
```
asinghvi17: What's this do?

src/attributes.jl
```diff
@@ -0,0 +1,195 @@
+"""
+Main structure for holding attributes, for theming plots etc!
+"""
+struct Attributes
+    # name, for better error messages!
+    name::String
+    # We dont have one node per value anymore, but instead one node
+    # that gets triggered on any setindex!, or whenever an input attribute node changes
+    # This makes it easier to layer Observable{Attributes}()
+    on_change::Node{Pair{Symbol, Any}}
```
asinghvi17: With this, would a re-render be forced whenever you update `scene.color`?  I'm not sure how this would work, would appreciate some clarification here.

src/attributes.jl
```diff
@@ -0,0 +1,195 @@
...
+    default_values::Dict{Symbol, Any}
+
+    function Attributes(name::String, default_values::Dict{Symbol, Any}, from_user::Dict{Symbol, Any})
+        on_change = Node{Pair{Symbol, Any}}()
+        supported_fields = Set(keys(default_values))
+        return new(name, on_change, supported_fields, from_user,
+                   Dict{Symbol, Any}(), default_values)
+    end
+end
+
```
asinghvi17:
```suggestion

Attributes(default_values::Dict{Symbol, Any}, from_user::Dict{Symbol, Any}) = Attributes(gensym("attrs"), default_values, from_user)

```

src/attributes.jl
```diff
@@ -0,0 +1,195 @@
...
+        merge_attributes!(target, elem)
+    end
+    return target
+end
+
+Base.merge(target::Attributes, args::Attributes...) = merge!(copy(target), args...)
+
+@generated hasfield(x::T, ::Val{key}) where {T, key} = :($(key in fieldnames(T)))
+
+function Base.getproperty(x::T, key::Symbol) where T <: Union{Attributes, Transformable}
```
asinghvi17: We should also define `propertynames`!

src/attributes.jl
```diff
@@ -0,0 +1,195 @@
...
+    supported_fields::Set{Symbol}
+    # The attributes given at construction time, taking the highest priority
+    from_user::Dict{Symbol, Any}
+    # attributes filled in by e.g. a theme, or other processes in the pipeline
+    from_theme::Dict{Symbol, Any}
+    # The "global" default values for this specific attribute instance
+    # Will be immutabe (maybe not a Dict then?) and shared between all similar objects
+    default_values::Dict{Symbol, Any}
+
+    function Attributes(name::String, default_values::Dict{Symbol, Any}, from_user::Dict{Symbol, Any})
```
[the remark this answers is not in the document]
SimonDanisch: That'd would defeat its purpose.. I'd use `""` if someone doesn't want to name it..
This is for having errors a la `scatter doesn't support field colur`

src/attributes.jl
SimonDanisch: That never triggered a rerender in opengl, and still wont ;) But I guess will trigger one in other frameworks... But its essentially not different from the Attribute type before, just an optimized implementation!

src/attributes.jl
```diff
@@ -1,13 +1,14 @@
+
 """
-Main structure for holding attributes, for theming plots etc!
+Main structure for holding attributes in e.g. plots!
 """
-struct Attributes
+struct ObservableAttributes
```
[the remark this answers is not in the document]
SimonDanisch: I couldn't get it to replace the current `Attributes`, so I went for damage control and decided to integrate it more slowly

piever:
Not sure if this is the right place to comment, but here's one thing that came out (when playing with the "algebra of graphics" stuff). I think the new transform pipeline should be designed in such a way that it could be overloaded either based on the plot type for all arguments (already possible) or on the argument types for all plots (not easy right now, only `convert_arguments` hack).

I think this is handled well by for example https://github.com/JuliaGPU/Adapt.jl (in a completely different setting). They split `adapt` in `adapt_structure` and `adapt_storage`, which you should overload for wrapper type or innermost type respectively.

SimonDanisch:
Can you give me 2 concrete examples so that I can add them to the tests for the new  pipeline?

src/theming.jl
```diff
@@ -1,3 +1,53 @@
...
+        # ccall((:GetSystemMetricsForDpi, :user32), Cint, (Cint, Cuint), 0, ccall((:GetDpiForSystem, :user32), Cuint, ()))
+        # ccall((:GetSystemMetrics, :user32), Cint, (Cint,), 17)
+        dc = ccall((:GetDC, :user32), Ptr{Cvoid}, (Ptr{Cvoid},), C_NULL)
+        ntuple(2) do i
+            Int(ccall((:GetDeviceCaps, :gdi32), Cint, (Ptr{Cvoid}, Cint), dc, (2 - i) + 117))
+        end
+    end
+elseif Sys.isapple()
+    function _primary_resolution()
+        s = read(pipeline(`system_profiler SPDisplaysDataType`, `grep Resolution`)) |> String
```
asinghvi17:
```suggestion
        s = read(pipeline(`system_profiler SPDisplaysDataType`, `grep Resolution`), String)

**Comment by piever** [comment]
Mhm, actually by exploring a bit more, I think it was a matter of finding the right method, and it already works well (I'll try to see if it's powerful enough to support grammar of graphics). I think I had a misconception on how this is done in Makie. So, one example (overload on the arguments) could be for the Measurements package.

```julia
julia> using Measurements: ±, value, uncertainty, Measurement

julia> using AbstractPlotting

julia> using AbstractPlotting: SceneLike, PlotFunc

julia> function AbstractPlotting.plot!(scn::SceneLike, P::PlotFunc, attr::Attributes, v::Vector{<:Measurement})
           val = value.(v)
           u = uncertainty.(v)
           plot!(scn, P, attr, val)
           plot!(scn, Band, attr, 1:length(v), val - u, val + u)
           return scn
       end
```

And now, if I define a custom recipe with `@recipe`, say 

```julia
julia> using Colors

julia> @recipe(RedPlot, x) do scene
           default_theme(scene)
       end

julia> AbstractPlotting.plot!(plot::RedPlot) = lines!(plot, plot[:x], color = colorant"red")
```

I can already do `redplot(rand(10) .± 0.1)`. From what I understand, in the refactor this type of approach will be the default, so we need to make sure that this example remains possible.

SimonDanisch:
Changes in https://github.com/JuliaPlots/AbstractPlotting.jl/pull/426

PR #362, opened 2020-03-15
remove old light attribute

ffreyer:
Related to https://github.com/JuliaPlots/GLMakie.jl/pull/75

### DIFF
--- a/src/documentation/documentation.jl
+++ b/src/documentation/documentation.jl
@@ -112,7 +112,7 @@ function help_attributes(io::IO, Typ::Type{T}; extended = false) where T <: Abst
     attributes = default_theme(nothing, Typ)
 
     # manually filter out some attributes
-    filter_keys = Symbol.([:fxaa, :model, :transformation, :light])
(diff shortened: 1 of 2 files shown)


ffreyer:
Quick summary for https://github.com/JuliaPlots/GLMakie.jl/pull/75:

- `light` changed to `ambient::Vec3f0`, `diffuse:.Vec3f0`, `specular::Vec3f0`, `shininess::Float32` and `lightposition::Vec3f0`
- now behaves as if a light source is placed at `lightposition`
- defaults set on GLMakie's side with much more ambient light
- removed secondary light source

SimonDanisch:
So, we shouldn't set any defaults in the backend..That GLMakie even has a pipeline for that is just because of historical reason and it will go any minute ;) 
Would be nice, if you could, instead of removing those here, actually add them here (e.g., so that we can also implement them in WGLMakie!). 
If not, I will soon test your PR out and make sure it passes tests, so I can do that at that point as well!

PR #366, opened 2020-03-17
Add GR's `peaks` function

asinghvi17:


asinghvi17:
All examples are failing, and I'm not sure why.  @SimonDanisch I think something broke GLMakie?

SimonDanisch:
Oh nice... I guess the master version of AbstractPlotting currently requires the master version of GLMakie

SimonDanisch:
Lets tag it!

asinghvi17:
Ah, is the new lighting stuff in?  Nice...we should write up a section in the backend page

asinghvi17:
I'll merge this, and then let's tag!

SimonDanisch:
GLMakie needs to be tagged first! It contains a non breaking change, to unbreak AbstractPlotting#master :P

asinghvi17:
Ah, that makes sense...we should put a constraint in Makie also, in that case.

PR #369, opened 2020-03-19
Prepare a release

asinghvi17:


asinghvi17:
argh, this pkg crap again

asinghvi17:
The orthographic camera does actually fail, but the rest of the examples look fine.
Refimage:
ref_image
Testimage:
test_image

SimonDanisch:
that looks fine :)

ffreyer:
Probably a result of the updates to light calculations. Running the example with
```julia
scene = mesh(merge(meshes), ambient=Vec3f0(0.03), specular=Vec3f0(0.3), diffuse=Vec3f0(0.9), lightposition=Vec3f0(20))
```
to restore the previous values gives

Screenshot from 2020-03-24 01-12-08

This shouldn't be/isn't exactly the same because the way `o_vertex ` is calculated in the shader has changed.

asinghvi17:
@SimonDanisch should we go ahead and tag AbstractPlotting + GLMakie?

piever:
StatsMakie master branch at the moment relies on some features that only exist on AbstractPlotting master, it'd be great to tag here so that I can add a lower bound and tag there as well.

asinghvi17:
I'm re-running Gitlab now; if it passes, I'll releaase.

PR #374, opened 2020-03-26
[WIP - don't merge] better text layouting

jkrumbiegel:
this fixes wonky glyphs when using annotations! but not simple text! for some reason. also in GLMakie it seems to introduce some slight bugs with minus dashes of tick labels. also I thought that the conversion factor for freetype units was 64, but it seems I have to additionally divide the resulting glyph positions by 4 in order to make it work


src/layouting/layouting.jl
```diff
@@ -21,16 +21,96 @@ function layout_text(
...
+
     return positions, scales
 end
 
+
+function glyph_positions(str::AbstractString, font, fontscale, halign, valign; lineheight_factor = 1.0, justification = 0.0)
+
+    # this is a countermeasure against Cairo messing with FreeType font pixel sizes
+    # when drawing. We reset them every time which is hacky but seems to work
+    FreeTypeAbstraction.FreeType.FT_Set_Pixel_Sizes(font, 64, 64)
```
SimonDanisch, line 50: Shouldn't Cairo hacks go into CairoMakie?

src/layouting/layouting.jl
```diff
@@ -21,16 +21,96 @@ function layout_text(
...
+    FreeTypeAbstraction.FreeType.FT_Set_Transform(font, C_NULL, C_NULL)
+    
+
+    # make lineheight a multiple of font's M height
+    lineheight = inkheight(FreeTypeAbstraction.internal_get_extent(font, 'M')) * lineheight_factor
+
+    lines = split(str, "\n")
+    extents = [[FreeTypeAbstraction.internal_get_extent(font, c) for c in l] for l in lines]
```
SimonDanisch, line 58: The caching turned out to be pretty important for decent performance, so if the divide into normal units doesn't work here, we should fix that in FreeTypeAbstraction

src/layouting/layouting.jl
```diff
@@ -21,16 +21,96 @@ function layout_text(
...
+
+    xkernings = [[FreeTypeAbstraction.kerning(c1, c2, font)[1]
+        for (c1, c2) in zip(chop(l, head = 0, tail = 1), chop(l, head = 1, tail = 0))]
+            for l in lines]
+
+    # add or subtract kernings?
+    xs = [cumsum([0; hadvance.(extgroup[1:end-1]) .+ kerngroup]) for (extgroup, kerngroup) in zip(extents, xkernings)]
+    [hadvance.(extgroup[1:end-1]) for extgroup in extents]
```
SimonDanisch, line 66: This line seems unused?

SimonDanisch:
Closed in favor of: https://github.com/JuliaPlots/AbstractPlotting.jl/pull/395

src/layouting/layouting.jl
jkrumbiegel: in principle yes but cairomakie affects freetype state.. so the countermeasure has to go there too. but maybe switching to a workflow without FT_Set_Pixel_Sizes works. It seems that what Cairo resets the fonts to is the standard size matching `font.units_per_EM`

src/layouting/layouting.jl
jkrumbiegel: this is a side effect of me trying to fix the changing font pixel sizes, so once we fix that, this workaround isn't needed and we can cache again

src/layouting/layouting.jl
jkrumbiegel: true, think I deleted that in the newer versions

PR #375, opened 2020-03-28
Pass kwargs to `backend_show` in `save`

asinghvi17:
I'm not sure whether this is the right way to do this, but it seems to make sense...

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -211,7 +211,7 @@ function FileIO.save(
 
     resolution !== size(scene) && resize!(scene, resolution)
     open(FileIO.filename(f), "w") do s
-        show(IOContext(s, merge((full_fidelity = true,), kwargs), format2mime(F), scene)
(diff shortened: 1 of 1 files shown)


SimonDanisch:
what's the use case? We should add tests + examples for this ;)

asinghvi17:
The original usecase for this was the ability to change the DPI when rendering PNGs in CairoMakie (you can read an in-memory SVG into Rsvg, which can then render to an upscaled PNG).

SimonDanisch:
Well this needs to be documented then...and best write out the keyword arguments ;)
Like it is right now, its impossible to know what this is for, why its there and how one can use it.

SimonDanisch:
Closing for now, feel free to reopen once we have some clear direction / docs for this!

asinghvi17:
@jkrumbiegel has a way to render arbitrary-DPI raster images in CairoMakie, so we could document this directly in the CairoMakie README.  This would also allow, e.g., metadata forwarding for PDF.

SimonDanisch:
Implemented by #407

PR #383, opened 2020-04-08
Add a compression factor option to `record`

asinghvi17:


asinghvi17:
Could someone test this, to see if it works correctly?

SimonDanisch:
Did you test it with a couple of examples?

asinghvi17:
I tested it, but am not sure if it works.  Will try on a really large example using CairoMakie.

asinghvi17:
Here's a zipfile with several videos.  For high resolutions, you can see the effect of compression.
https://drive.google.com/open?id=1FVITSTaI9kRUqmbqxX5RVSM7LTrKqNhd

PR #390, opened 2020-04-10
give poly default strokewidth of 1

jkrumbiegel:
If you want to give poly a stroke, you always have to set both strokecolor and strokewidth right now which is quite annoying. I think if you specify a color then a stroke in that color should be visible, then you can still tweak the linewidth but 1 is a good default. It could also be the other way around, but I think it's more natural to think about the color first.

This won't change any plot using only the defaults, because the effect of strokewidth 1 and color = transparent is still an invisible stroke.


mschauer:
Does the stroke width affect the interior in any way if the stroke is invisible by choice of alpha?

asinghvi17:
In general, no - the stroke is a `wireframe` over the poly `mesh`.

SimonDanisch:
Seems fair ;)

PR #398, opened 2020-04-19
two new selection functions

Datseris:
Please add `select_rectangle, select_line, select_point` to the "Atomic interaction functions" section of the Makie docs once you merge!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "AbstractPlotting"
 uuid = "537997a7-5e4e-5d89-9595-2241ea00577e"
-version = "0.10.1"
+version = "0.11.0"
 
 [deps]
 AbstractNumbers = "85c772de-338a-5e7f-b815-41e76c26ac1f"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "AbstractPlotting"
 uuid = "537997a7-5e4e-5d89-9595-2241ea00577e"
-version = "0.11.0"
+version = "0.10.2"
 
 [deps]
 AbstractNumbers = "85c772de-338a-5e7f-b815-41e76c26ac1f"
--- a/src/interaction/interactive_api.jl
+++ b/src/interaction/interactive_api.jl
@@ -1,5 +1,5 @@
 export mouseover, mouse_selection, mouseposition, hovered_scene
-export select_rectangle
+export select_rectangle, select_line, select_point
 
 
 """
@@ -124,7 +124,7 @@ hovered_scene() = error("hoevered_scene is not implemented yet.")
 
 """
     select_rectangle(scene; kwargs...) -> rect
-Interactively select a rectangle on a `scene` by clicking the left mouse button,
+Interactively select a rectangle on a 2D `scene` by clicking the left mouse button,
 dragging and then un-clicking. The function returns an **observable** `rect` whose
 value corresponds to the selected rectangle on the scene. In addition the function
 _plots_ the selected rectangle on the scene as the user clicks and moves the mouse
@@ -143,8 +143,8 @@ function select_rectangle(scene; kwargs...)
     rect_ret = Node(FRect(0, 0, 1, 1)) # returned rectangle
 
     # Create an initially hidden rectangle
-    rect_vis = lines!(
-        scene, rect, raw = true, visible = false, kwargs...,
+    plotted_rect = lines!(
+        scene, rect, raw = true, visible = false, color = RGBAf0(0.1, 0.1, 0.8, 0.5), kwargs...,
     )[end] # Why do I have to do [end] ?
 
     on(events(scene).mousedrag) do drag
@@ -152,7 +152,7 @@ function select_rectangle(scene; kwargs...)
             mp = mouseposition(scene)
             if drag == Mouse.down
                 waspressed[] = true
-                rect_vis[:visible] = true # start displaying
+                plotted_rect[:visible] = true # start displaying
                 rect[] = FRect(mp, 0.0, 0.0)
             elseif drag == Mouse.pressed
                 mini = minimum(rect[])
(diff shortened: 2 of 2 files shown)


src/interaction/interactive_api.jl
```diff
@@ -143,16 +143,16 @@ function select_rectangle(scene; kwargs...)
     rect_ret = Node(FRect(0, 0, 1, 1)) # returned rectangle
 
     # Create an initially hidden rectangle
-    rect_vis = lines!(
-        scene, rect, raw = true, visible = false, kwargs...,
+    plotted_rect = lines!(
+        scene, rect, raw = true, visible = false, color = RGBAf0(0.1, 0.1, 0.8, 0.5), kwargs...,
```
asinghvi17, line 147: Let's make this themable!

src/interaction/interactive_api.jl
```diff
@@ -168,9 +168,107 @@ function select_rectangle(scene; kwargs...)
...
+The `kwargs...` are propagated into `lines!` which plots the selected line.
+"""
+function select_line(scene; kwargs...)
+    key = Mouse.left
+    waspressed = Node(false)
+    line = Node([Point2f0(0,0), Point2f0(1,1)])
+    line_ret = Node([Point2f0(0,0), Point2f0(1,1)])
+    # Create an initially hidden  arrow
+    plotted_line = lines!(
+        scene, line; visible = false, color = RGBAf0(0.1, 0.1, 0.8, 0.5),
```
asinghvi17, line 197: Same as above - this should be a kwarg.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "AbstractPlotting"
 uuid = "537997a7-5e4e-5d89-9595-2241ea00577e"
-version = "0.10.1"
+version = "0.11.0"
```
asinghvi17, line 3: If this doesn't break anything, I'd rather it be a patch version bump...

src/interaction/interactive_api.jl
```diff
@@ -168,9 +168,107 @@ function select_rectangle(scene; kwargs...)
...
+_plots_ the line on the scene as the user clicks and moves the mouse
+around. When the button is not clicked any more, the plotted line disappears.
+
+The value of the returned line is updated **only** when the user un-clicks.
+
+The `kwargs...` are propagated into `lines!` which plots the selected line.
+"""
+function select_line(scene; kwargs...)
+    key = Mouse.left
+    waspressed = Node(false)
```
asinghvi17, line 192: If you never lift on this, it could also be a `Ref`, although I don't know exactly how much overhead that shaves off.
Also, it'd be nice to use `Observable` instead of `Node`, so we can start to move away from that.

src/interaction/interactive_api.jl
```diff
@@ -168,9 +168,107 @@ function select_rectangle(scene; kwargs...)
...
+"""
+function select_point(scene; kwargs...)
+    key = Mouse.left
+    pmarker = Circle(Point2f0(0, 0), Float32(1))
+    waspressed = Node(false)
+    point = Node([Point2f0(0,0)])
+    point_ret = Node([Point2f0(0,0)])
+    # Create an initially hidden  arrow
+    plotted_point = scatter!(
+        scene, point; visible = false, marker = pmarker, markersize = 20px,
```
asinghvi17, line 247: Markersize should also ideally be themable.

src/interaction/interactive_api.jl
Datseris: I don't understand what you are saying, please be explicit.

src/interaction/interactive_api.jl
Datseris: I disagree: The healthy way to do it is to load up Atom and do "Find all Replace all" on `Node`. Doing it one by one seems insane to me.

src/interaction/interactive_api.jl
asinghvi17: I just mean that it should be passable as a kwarg, so that the color isn't hardcoded.

src/interaction/interactive_api.jl
Datseris: once again I don't understand. Please be explicit.

src/interaction/interactive_api.jl
Datseris: The color isn't hardcoded, this is not how keywords work. If there is a keyword `color` in the propagated kwargs, it overwrites the "hardcoded".

src/interaction/interactive_api.jl
asinghvi17: Fair enough, there aren't any API incompatibilities (since it's the same API) so that should be a 5-minute job.

Incidentally, what are your thoughts on `Obs` as a shorthand for `Observable`?

src/interaction/interactive_api.jl
asinghvi17: Same thing as above - `marker` and `markersize` should be passed as kwargs by the user.

src/interaction/interactive_api.jl
asinghvi17: Huh, didn't know about that.  Pretty neat!

src/interaction/interactive_api.jl
Datseris: I think it is great.

src/interaction/interactive_api.jl
Datseris: What are you talking about? That's not how keyword propagation works, please read my other comment.

Datseris: Ah okay, you saw it.

asinghvi17: Yep, sorry - forgot to mark as resolved.  I was unaware that kwargs would merge.

src/interaction/interactive_api.jl
```diff
@@ -214,7 +215,9 @@ function select_line(scene; kwargs...)
         else
             if drag == Mouse.up && waspressed[] # User has selected the rectangle
                 waspressed[] = false
-                line_ret[] = copy(line[])
+                if sum(abs2.(line[][1] - line[][2])) > 0
```
asinghvi17, line 218:
```suggestion
                if line[][1] != line[][2]

Datseris:
I need this to release the interactive billiard app I've been developing so I'd appreciate it if a version with these functions is tagged soon.

PR #400, opened 2020-04-19
Add a dispatch for `(colormap::Symbol, alpha::Real)`

### LINKED ISSUE #399: [FR] Add "alpha" keyword to `to_colormap`.
Will make `AbstractPlotting.to_colormap(:viridis, N)` very convenient!

asinghvi17:
Fixes #399

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -777,6 +777,11 @@ function convert_attribute(cs::Union{Tuple, Pair}, ::key"colormap", n::Int=2)
     return to_colormap([to_color.(cs)...], n)
 end
 
+function convert_attribute(cs::Tuple{Union{Symbol, AbstractString}, Real}, ::key"colormap", n::Int=30)
+    return RGBAf0.(to_colormap(cs[1]), cs[2]) # We need to rework this to conform to the backend interface.
+end
+
+
 to_colormap(x, n::Integer) = convert_attribute(x, key"colormap"(), n)
 
 """
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -781,7 +781,7 @@ function convert_attribute(cs::Tuple{<: Union{Symbol, AbstractString}, Real}, ::
     return RGBAf0.(to_colormap(cs[1]), cs[2]) # We need to rework this to conform to the backend interface.
 end
 
-Tuple{Union{Symbol, AbstractString}, Real, Int}
+function convert_attribute(cs::NamedTuple{(:colormap, :alpha, :n), Tuple{Union{Symbol, AbstractString}, Real, Int}}, ::key"colormap")
     return RGBAf0.(to_colormap(cs.colormap, cs.n), cs.alpha)
 end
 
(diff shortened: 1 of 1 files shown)


src/conversions.jl
```diff
@@ -777,6 +777,11 @@ function convert_attribute(cs::Union{Tuple, Pair}, ::key"colormap", n::Int=2)
     return to_colormap([to_color.(cs)...], n)
 end
 
+function convert_attribute(cs::Tuple{Union{Symbol, AbstractString}, Real}, ::key"colormap", n::Int=30)
+    return RGBAf0.(to_colormap(cs[1]), cs[2]) # We need to rework this to conform to the backend interface.
```
SimonDanisch, line 781: ># We need to rework this to conform to the backend interface.

What needs to be reworked here?

Btw, while we're at it, maybe we can add also one for n? Maybe a named tuple?

asinghvi17: So my primary concern with the vector of rgb solution we currently have, is that we also have the sampler interface, and PlotUtils also has their own (new) interface for scaling etc.

It just feels like a lot of duplicated code when we should standardize on one implementation.

SimonDanisch: Yeah... I will need to clean this up soon ;) This is my second priority after merging the font layout!

src/conversions.jl
```diff
@@ -777,6 +777,15 @@ function convert_attribute(cs::Union{Tuple, Pair}, ::key"colormap", n::Int=2)
     return to_colormap([to_color.(cs)...], n)
 end
 
+function convert_attribute(cs::Tuple{<: Union{Symbol, AbstractString}, Real}, ::key"colormap", n::Int=30)
+    return RGBAf0.(to_colormap(cs[1]), cs[2]) # We need to rework this to conform to the backend interface.
+end
+
+function convert_attribute(cs::NamedTuple{(:colormap, :alpha, :n), Tuple{<: Union{Symbol, AbstractString}, <: Real, Int}}, ::key"colormap")
```
asinghvi17, line 784:
```suggestion
Tuple{Union{Symbol, AbstractString}, Real, Int}

Comment (asinghvi17) [reply src/conversions.jl]:
```suggestion
function convert_attribute(cs::NamedTuple{(:colormap, :alpha, :n), Tuple{Union{Symbol, AbstractString}, Real, Int}}, ::key"colormap")

File: src/conversions.jl [hunk 3]
```diff
@@ -777,6 +777,15 @@ function convert_attribute(cs::Union{Tuple, Pair}, ::key"colormap", n::Int=2)
     return to_colormap([to_color.(cs)...], n)
 end
 
+function convert_attribute(cs::Tuple{<: Union{Symbol, AbstractString}, Real}, ::key"colormap", n::Int=30)
+    return RGBAf0.(to_colormap(cs[1]), cs[2]) # We need to rework this to conform to the backend interface.
+end
+
+Tuple{Union{Symbol, AbstractString}, Real, Int}
```
asinghvi17, line 784:
```suggestion
function convert_attribute(cs::NamedTuple{(:colormap, :alpha, :n), Tuple{Union{Symbol, AbstractString}, Real, Int}}, ::key"colormap")

PR #401, opened 2020-04-19
 New text layout

SimonDanisch:


SimonDanisch:
Needs:
https://github.com/JuliaGeometry/GeometryBasics.jl/pull/25
https://github.com/JuliaGraphics/FreeTypeAbstraction.jl/pull/45

src/layouting/boundingbox.jl
```diff
@@ -172,14 +172,14 @@ function boundingbox(
                 inkboundingbox(FreeTypeAbstraction.get_extent(font, c))
             end
 
-            scaled_bb = bb_unitspace * scale
+            scaled_bb = rectmult(bb_unitspace, scale)
```
SimonDanisch, line 175: I moved this to GeometryBasics where it should live! You need the newest version for it though

SimonDanisch: So would be nice to revert this ;)

src/utilities/texture_atlas.jl
```diff
@@ -106,7 +106,7 @@ begin
             end
         end
         atlas = TextureAtlas()
-        @info("Caching fonts, this may take a while. Needed only on first run!")
+        @info("Makie/AbstractPlotting Caching fonts, this may take a while. Needed only on first run!")
```
asinghvi17, line 109:
```suggestion
        @info("Makie/AbstractPlotting is caching fonts, this may take a while. Needed only on first run!")

PR #407, opened 2020-04-22
add CairoMakie kwargs and fix bug with corrupt existing files

jkrumbiegel:
the bug was that if there was an existing corrupt test.png for example, the fileio type parameter would be :UNKNOWN, as read out from the corrupt / nonexistent magic bytes. We don't care what is going on with any existing file, therefore we delete it and determine the filetype parameter only from the file name.

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -193,6 +193,7 @@ format2mime(::Type{FileIO.format"TEX"})  = MIME("application/x-tex")
 format2mime(::Type{FileIO.format"EPS"})  = MIME("application/postscript")
 format2mime(::Type{FileIO.format"HTML"}) = MIME("text/html")
 
+filetype(::FileIO.File{F}) where F = F
 # Allow format to be overridden with first argument
 """
     FileIO.save(filename, scene; resolution = size(scene))
@@ -205,13 +206,32 @@ Allowable formats depend on the backend;
 Resolution can be specified, via `save("path", scene, resolution = (1000, 1000))`!
 """
 function FileIO.save(
-        f::FileIO.File{F}, scene::Scene;
-        resolution = size(scene)
-    ) where F
+        f::FileIO.File, scene::Scene;
+        resolution = size(scene), kwargs...
+    )
 
     resolution !== size(scene) && resize!(scene, resolution)
-    open(FileIO.filename(f), "w") do s
-        show(IOContext(s, :full_fidelity => true), format2mime(F), scene)
+
+    # Delete previous file if it exists and query only the file string for type.
+    # We overwrite existing files anyway, so this doesn't change the behavior.
+    # But otherwise we could get a filetype :UNKOWN from a corrupt existing file
+    # (from an error during save, e.g.), therefore we don't want to rely on the
+    # type readout from an existing file.
+    filename = FileIO.filename(f)
+    isfile(filename) && rm(filename)
+    # query the filetype only from the file extension
+    F = filetype(FileIO.query(filename))
+
+    # we can pass arbitrary keywords to the backend saving functions
+    # which they can retrieve using `get(::IOContext, :kw, default_value)`
+    # an example would be a resolution scaling factor for CairoMakie's png backend
+    kwarg_pairs = pairs((full_fidelity = true, kwargs...))
+
+    open(filename, "w") do s
+        show(
+            IOContext(s, kwarg_pairs...),
+            format2mime(F),
+            scene)
     end
 end
 
--- a/src/display.jl
+++ b/src/display.jl
@@ -214,7 +214,7 @@ function FileIO.save(
 
     # Delete previous file if it exists and query only the file string for type.
     # We overwrite existing files anyway, so this doesn't change the behavior.
-    # But otherwise we could get a filetype :UNKOWN from a corrupt existing file
+    # But otherwise we could get a filetype :UNKNOWN from a corrupt existing file
     # (from an error during save, e.g.), therefore we don't want to rely on the
     # type readout from an existing file.
     filename = FileIO.filename(f)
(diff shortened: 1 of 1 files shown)


src/display.jl
```diff
@@ -205,13 +206,32 @@ Allowable formats depend on the backend;
...
+        resolution = size(scene), kwargs...
+    )
 
     resolution !== size(scene) && resize!(scene, resolution)
-    open(FileIO.filename(f), "w") do s
-        show(IOContext(s, :full_fidelity => true), format2mime(F), scene)
+
+    # Delete previous file if it exists and query only the file string for type.
+    # We overwrite existing files anyway, so this doesn't change the behavior.
+    # But otherwise we could get a filetype :UNKOWN from a corrupt existing file
```
asinghvi17, line 217:
```suggestion
    # But otherwise we could get a filetype :UNKNOWN from a corrupt existing file

File: src/display.jl [hunk 2]
```diff
@@ -205,13 +206,32 @@ Allowable formats depend on the backend;
 Resolution can be specified, via `save("path", scene, resolution = (1000, 1000))`!
 """
 function FileIO.save(
-        f::FileIO.File{F}, scene::Scene;
-        resolution = size(scene)
-    ) where F
+        f::FileIO.File, scene::Scene;
+        resolution = size(scene), kwargs...
```
SimonDanisch, line 210: Can we write out the expected kwargs and document them?

jkrumbiegel: they differ for each backend, though, so they would end up in those backends? for png in Cairo we need `px_per_unit` and for svg/pdf `pt_per_unit`

jkrumbiegel: `unit` then refers to the scene "pixels" we currently have

SimonDanisch: Well, even more important to document them here!
This is still the function people call and will call help/autocomplete on, if they dont understand how to use it, no?
Also, cant we just use pt?
`In typography, the point is the smallest unit of measure` so for a pixel based backend, this means pt==pixel, no?

SimonDanisch: Btw, I have a pretty big pet peeve with `kw...` if its not 100% used for a function that just passes them through and has no way to actually write them out. 
Its so hard find out with multiple dispatch, where those keyword arguments actually land and therefore really hard to use them and figure out what is allowed and what not...
You will need a debugger to actually find where those keyword arguments are ending up - which gets even harder, when they're transferred to the IO object :D

jkrumbiegel: I agree, but do you have a better alternative? We need arbitrary keywords being able to end up at our backend functions, and different keywords are needed for different backends.

jkrumbiegel: And no, we can't call the scene units pt unless we completely lock that conversion in. But in GLMakie one unit is obviously a pixel right now, not a pt. And in pdf / svg pt has a quite specific meaning.

SimonDanisch: >But in GLMakie one unit is obviously a pixel right now, not a pt

I interpreted that pt is the smallest unit of the drawing "backend" which is pixels for GLMakie - or did I misinterpret the meaning of pt from wikipedia? :D 

>We need arbitrary keywords being able to end up at our backend functions

No, we need to have a well defined set of keyword arguments that end up in different backends... This is exactly the point where we should document those differences for different backends :)

SimonDanisch: If you think the pt point of mine is just too confusing because of pre-existing concepts, or because i misunderstood, I won't insist on this :D 
I was just looking at the definition and thought, that's exactly what we would want.

jkrumbiegel: A point is a physical measure, it's generally 1/72th of an inch. I wouldn't tie myself to that but instead use an arbitrary "base unit" that is then converted to a meaningful base unit for the backend. Even if that conversion just ends up as 1:1. In CairoMakie it's important to differentiate what it means to have a linewidth of 1, in pdf it's 1pt without any intervention and in png it's 1px.

jkrumbiegel: But it's easy to imagine different keywords to configure a Cairo png backend vs a Cairo svg backend, vs a GR backend etc no?

I have an idea, maybe each backend offers objects for this called PngConfig, SvgConfig, etc and the user passes one of those to the save function. Then those configs could have a limited set of allowed keywords

SimonDanisch: ok fair enough....

jkrumbiegel:
Another idea, instead of a special type for dispatch, can't we just use one keyword `config` which can be a named tuple or dict etc to overwrite a specific backend's settings? It's much easier to check that the content of that one object doesn't contain any unregistered keywords.

```julia
save("plot.png", scene, config = (px_per_unit = 2, antialiasing = :best, not_supported = 123...))


# then in the backend

config = get!(iocontext, :config, NamedTuple())

for key in config
	if !(key in allowed_keys)
        error()
    end
end
```

SimonDanisch:
Sounds good to me, as long as its documented here ;)

jkrumbiegel:
The `config` or `options` or whatever keyword will be documented here, but the actual options have to be documented by the backends. Maybe we can make a helper method to discover the options for a given backend.

jkrumbiegel:
How do you like it now?

src/display.jl
```diff
@@ -193,25 +193,59 @@ format2mime(::Type{FileIO.format"TEX"})  = MIME("application/x-tex")
...
+    F = filetype(FileIO.query(filename))
+
+    kwarg_pairs = pairs((full_fidelity = true, pt_per_unit = pt_per_unit,
+        px_per_unit = px_per_unit))
+
+    open(filename, "w") do s
+        show(
+            IOContext(s, kwarg_pairs...),
+            format2mime(F),
+            scene)
```
asinghvi17, line 248:
```suggestion
            scene
        )

SimonDanisch:
Looks great, thanks a lot for going through this!

PR #413, opened 2020-04-25
Allow linestyle = :solid

asinghvi17:
Plots also exposes this option, and it seems more intuitive than `nothing`.

@Datseris


Datseris:
Bump patch version as well? :)

asinghvi17:
I can't merge this until we fix CI, but will let you know when that happens.

PR #415, opened 2020-04-28
Fix Juno plotpane display in CairoMakie

asinghvi17:


src/display.jl
```diff
@@ -70,13 +70,14 @@ end
 function backend_show end
 
 for M in (MIME"text/plain", MIME)
-    @eval function Base.show(io::IO, m::$M, scene::Scene)
+    @eval AbstractPlotting function Base.show(io::IO, m::$M, scene::Scene)
```
asinghvi17, line 73:
```suggestion
    @eval function Base.show(io::IO, m::$M, scene::Scene)

**Comment by asinghvi17** [comment]
Reviewed on Slack by @jkrumbiegel

**Comment by jkrumbiegel** [comment]
The only thing I am still wondering about is if it's better to reduce the scene resolution by 0.75, instead of using the scaling multiplier. Because using the scaling multiplier will affect font sizes, maybe people expect that 12units will in the default case equal 12pt in the display, not 9pt like with the scaling

**Comment by jkrumbiegel** [comment]
```julia
res = get(io, :juno_plotsize, nothing)
if !isnothing(res)
    res = round.(Int, res .* 0.75)
end

res !== nothing && resize!(scene, res...)
ioc = IOContext(io, :full_fidelity => true, :pt_per_unit => get(io, :pt_per_unit, 1.0), :px_per_unit => get(io, :px_per_unit, 1.0))
```

PR #417, opened 2020-04-28
Changes for PlotUtils v1.0 compat

asinghvi17:


asinghvi17:
It works on my machine and will probably work when released.  CI is just being annoying.

PR #420, opened 2020-04-30
Use xvfb and remove xorg tag

asinghvi17:


SimonDanisch:
We now have OpenGL on gh-actions

PR #421, opened 2020-05-01
WIP: add highclip and lowclip to heatmap

jkrumbiegel:
defaulting to nothing, which means that the old behavior of continuing with the extreme colors is preserved


jkrumbiegel:
together with https://github.com/JuliaPlots/GLMakie.jl/pull/100

PR #422, opened 2020-05-01
add support for transform_func

SimonDanisch:


asinghvi17:
Would this handle transforms which depend on both x and y?  E.g. transform(x, y) = (y*x, y-x) or something

src/basic_recipes/axis.jl
```diff
@@ -538,6 +538,8 @@ end
 function plot!(scene::SceneLike, ::Type{<: Axis2D}, attributes::Attributes, args...)
     # create "empty" plot type - empty meaning containing no plots, just attributes + arguments
     cplot = Axis2D(scene, attributes, args)
+    # Disable any non linear transform for the axis plot!
```
asinghvi17, line 541: I suppose we should override these with different axis implementations?  For example, a generic axis which works across all transforms only requires you to densely sample your lines, such that they look smooth when plotted.

src/basic_recipes/basic_recipes.jl
```diff
@@ -1192,24 +1192,17 @@ function plot!(p::Spy)
...
-            BackendModule = parentmodule(typeof(AbstractPlotting.current_backend[]))
-            if nameof(BackendModule) == :GLMakie
-                BackendModule.FastPixel()
-            else
-                :rect
-            end
-        else
-            x
-        end
+        return x === automatic ? FastPixel() : x
```
asinghvi17, line 1203: I wonder if FastPixel should take an alpha modifier, so that you can use it to show not just the point but also the density of data within a pixel rect.

SimonDanisch: that can be done via the color, no?

asinghvi17: I'm not sure how it works.  If I have 2 points in my data which lie within the same pixel rectangle, and `color = (:red, .1)`, is the new alpha of that point now .2?

SimonDanisch: Should be!

PR #424, opened 2020-05-01
add SSAO attributes

ffreyer:
See https://github.com/JuliaPlots/GLMakie.jl/pull/97

- [x] done (if the attribute names are ok)


src/theming.jl
```diff
@@ -47,7 +47,7 @@ const minimal_default = Attributes(
     padding = Vec3f0(0.05),
     raw = false,
     SSAO = Attributes(
-        enable = Node(false),
+        # enable = Node(false),
         bias = Node(0.025f0),       # z threshhold for occlusion
```
asinghvi17, line 51: These should be automatically converted to nodes, so you shouldn't have explicitly declare it.

ffreyer: fixed it

PR #426, opened 2020-05-05
Refactor dict interface

SimonDanisch:


src/AbstractPlotting.jl
```diff
@@ -33,9 +31,20 @@ end
 using .ContoursHygiene
 const Contours = ContoursHygiene.Contour
 
+const RealVector{T} = AbstractVector{T} where T <: Number
+const Node = Observable# For now, we use Reactive.Signal as our Node type. This might change in the future
```
asinghvi17, line 35:
```suggestion
const Node = Observable # shorthand

File: src/attributes.jl [hunk 2]
```diff
@@ -0,0 +1,280 @@
...
+    on_change::Observable{Pair{Symbol, Any}}
+    # The supported fields, so we can throw an error, whenever fields are not supported
+    supported_fields::Set{Symbol}
+    # The attributes given at construction time, taking the highest priority
+    from_user::Dict{Symbol, Any}
+    # attributes filled in by e.g. a theme, or other processes in the pipeline
+    from_theme::Dict{Symbol, Any}
+    # The "global" default values for this specific attribute instance
+    # Will be immutabe (maybe not a Dict then?) and shared between all similar objects
+    default_values::Dict{Symbol, Any}
```
asinghvi17, line 20:
```suggestion
"""
    ObservableAttributes(; attrs...)
Main structure for holding attributes in e.g. plots!

## Extended help

### Fields
$(FIELDS)
"""
struct ObservableAttributes
    "name, for better error messages!"
    name::String
    """
    We dont have one node per value anymore, but instead one node
    that gets triggered on any setindex!, or whenever an input attribute node changes
    This makes it easier to layer Observable{ObservableAttributes}()
    """
    on_change::Observable{Pair{Symbol, Any}}
    "The supported fields, so we can throw an error, whenever fields are not supported"
    supported_fields::Set{Symbol}
    "The attributes given at construction time, taking the highest priority"
    from_user::Dict{Symbol, Any}
    "attributes filled in by e.g. a theme, or other processes in the pipeline"
    from_theme::Dict{Symbol, Any}
    """
    The "global" default values for this specific attribute instance
    Will be immutable (maybe not a Dict then?) and shared between all similar objects
    """
    default_values::Dict{Symbol, Any}
```

This integrates DocStringExtensions, so that you can see what the type is in the documentation and REPL.

src/attributes.jl
```diff
@@ -0,0 +1,280 @@
...
+    return (s[1] => x[s[1]], s[2])
+end
+
+function Base.copy(attributes::ObservableAttributes)
+    return ObservableAttributes(attributes)
+end
+
+Base.merge(target::ObservableAttributes, args::ObservableAttributes...) = merge!(copy(target), args...)
+
+function merge_attributes!(input::ObservableAttributes, theme::ObservableAttributes)
```
asinghvi17, line 226: why not overload `Base.merge!`?

src/attributes.jl
```diff
@@ -0,0 +1,280 @@
...
+        if !haskey(input, key)
+            input[key] = value
+        else
+            current_value = input[key]
+            if value isa ObservableAttributes && current_value isa ObservableAttributes
+                # if nested attribute, we merge recursively
+                merge_attributes!(current_value, value)
+            elseif value isa ObservableAttributes || current_value isa ObservableAttributes
+                error("""
+                Type missmatch while merging plot attributes with theme for key: $(key).
```
asinghvi17, line 237: ``suggestion
                Type mismatch while merging plot attributes with theme for key: $(key).

src/dictlike.jl
```diff
@@ -0,0 +1,248 @@
...
+const SceneLike = Union{AbstractScene, ScenePlot}
+
+"""
+Main structure for holding attributes, for theming plots etc!
+Will turn all values into nodes, so that they can be updated.
+"""
+struct Attributes
+    attributes::Dict{Symbol, Node}
+end
+const Theme = Attributes
```
asinghvi17, line 22: Is this redundant now with ObservableAttributes?

src/theming.jl
```diff
@@ -1,3 +1,53 @@
+if Sys.iswindows()
```
asinghvi17, line 1:
```suggestion
# This is a platform-specific way of getting the resolution
# of the current monitor, since each platform offers its own API.
if Sys.iswindows()
```

src/units.jl
```diff
@@ -1,3 +1,23 @@
...
+# abstract type DisplaySpace end
+# #...
+# struct Log <: DataSpace end
+# struct Polar <: DataSpace end
+# struct GeoProjection <: DataSpace end
+# # and more!
+# #...
+# struct Pixel <: DisplaySpace end
+# struct DPI <: DisplaySpace end
+# struct DIP <: DisplaySpace end
```
asinghvi17, line 17: I guess this wouldn't work with mixed spaces, as per the conversation in Slack

test/simple.jl
```diff
@@ -63,7 +64,14 @@ surface(args_2d..., color=rand(size(data_2d)...))
...
 surface(args_2d..., shading=false); wireframe!(args_2d..., linewidth=0.5)
-
+surface(1:30, 1:31, rand(30, 31))
+n = 20
+θ = [0;(0.5:n-0.5)/n;1]
+φ = [(0:2n-2)*2/(2n-1);2]
+x = [cospi(φ)*sinpi(θ) for θ in θ, φ in φ]
+y = [sinpi(φ)*sinpi(θ) for θ in θ, φ in φ]
+z = [cospi(θ) for θ in θ, φ in φ]
+surface(x, y, z)
```
asinghvi17, line 74: add to gallery?

asinghvi17:
With this refactor, could you still lift on individual attributes?  We may not want to rerun expensive pipelines, like streamplot, just for a colormap change.

PR #427, opened 2020-05-06
add x direction bar plot

jkrumbiegel:
I think direction :x or :y is a good interface in general for these kinds of plots.

One thing that still annoys me about barplots, is that plotting giving the edges and values is often more sensible, because the edges encode midpoints and widths. I wonder if we could do logic where if the first vector is one longer than the second one, automatic width calculation is more intelligent than this 0.8 default that we have here. This currently doesn't work because the PointBased conversion trait doesn't accept unequal length vectors

### DIFF
--- a/src/basic_recipes/basic_recipes.jl
+++ b/src/basic_recipes/basic_recipes.jl
@@ -489,7 +489,8 @@ $(ATTRIBUTES)
         marker = Rect,
         strokewidth = 0,
         strokecolor = :white,
-        width = automatic
+        width = automatic,
+        direction = :y,
     )
 end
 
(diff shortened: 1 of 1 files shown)


asinghvi17:
That would make sense, I think.  We could override the point-based conversion for barplot, while still keeping the trait (since it's useful for more general stuff).

At that point, we should probably also have a `calculated_attributes!` for barplot, to calculate the gap externally.

jkrumbiegel:
what needs to be done to merge this? for me this is one of those paper cuts I'd like to get rid of

SimonDanisch:
Just a ping to me, indicating its ready :D

PR #428, opened 2020-05-10
add pick by distance

ffreyer:
I came across
https://github.com/JuliaPlots/Makie.jl/issues/349
and thought I'd give it a try.

This pr adds a `range` argument to point picking functions. If it is used the closest plot/id within that (pixel) range is picked, rather than the plot/id at the mouseposition. 

see also: https://github.com/JuliaPlots/GLMakie.jl/pull/101


SimonDanisch:
Thanks, that's a good idea :)

Issue #276, opened 2020-01-19
camera zooming with orthographic projection type


jaakkor2:
In this example the zooming does not work out-of-plane, but moves the image in-plane.  I would expect to see similar zooming as in the `Perspective` option.
```julia
using Makie
using GLMakie
scene = wireframe(GLMakie.loadasset("cat.obj"))
cam = cameracontrols(scene)
cam.projectiontype[] = Makie.AbstractPlotting.Orthographic # not zooming
#cam.projectiontype[] = Makie.AbstractPlotting.Perspective # works as expected
update_cam!(scene, cam)
```

Edit: AbstractPlotting 0.9.17, GLMakie 0.0.13

jaakkor2:
Problematic lines are at
https://github.com/JuliaPlots/AbstractPlotting.jl/blob/8274bf4346a1ca43e1b1e880c9f2a37e9e89c61b/src/camera/camera3d.jl#L236-L238
if I comment those out, zooming in orthographic projection works as I would expect.

Those lines came from https://github.com/JuliaPlots/AbstractPlotting.jl/commit/08b63d29dd965e63b3cb128e8649be2bb5743d35, but the intent is unclear for me. Why would one shift `cam.lookat[]` when zooming? And why this is happening in orthographic projection, but not in perspective projection.

SimonDanisch:
Well, there are two ways to zoom, one is to change the field of view, the other to move the camera closer... To me it seemed like it makes sense, if moving the camera closer, to also move the look at... But maybe that's not true...
I just tried to set shift_lookat [here](https://github.com/JuliaPlots/AbstractPlotting.jl/blob/8274bf4346a1ca43e1b1e880c9f2a37e9e89c61b/src/camera/camera3d.jl#L82) to false and the camera doesn't seem to feel worse^^
So i'd be open to change that ;)

SimonDanisch:
Feel free to open a PR to merge that change ;)

PR #431, opened 2020-05-15
Fix 3d zooming

ffreyer:
Fixes https://github.com/JuliaPlots/Makie.jl/issues/583
Fixes https://github.com/JuliaPlots/AbstractPlotting.jl/issues/276
Might fix https://github.com/JuliaPlots/Makie.jl/issues/601

If `shift_lookat = true`, then the point under the cursor remains stationary when zooming - for both orthographic and perspective projection. With `shift_lookat = false` the camera still rotates. I'm not sure what the exact behavior is supposed to be in this case, so I went with whatever felt right.


ffreyer:
(This is ready to be merged if there are no complaints)

PR #434, opened 2020-05-23
add simple histogram recipe

jkrumbiegel:
may not be the most performant or optimized algorithm, but would be quite useful to have without taking on a StatsBase dependency


asinghvi17:
Doesn't StatsMakie already have this?

jkrumbiegel:
Yes, but statsmakie is a super heavy dependency. Arguably this is so basic it should just be a recipe in abstractplotting. I think the same about density plots

jkrumbiegel:
I get the grammar of graphics niche for statsmakie, but it shouldn't be necessary to do these simple plots. Same thing with MakieLayout, it shouldn't actually be necessary to add a package just for subplot support and usable colorbars / legends

piever:
Personally, I hope that StatsMakie can disappear at some point. I've been trying to move the Grammar of Graphics stuff to AlgebraOfGraphics, but for the other recipes I'm not too sure. I wonder whether it makes sense to ask again over at JuliaStats whether it would be possible for them to implement the recipes with RecipesBase and then consume those. AFAIU, the main objection to doing that in the past has been that this recipe system was to specific to a single plotting package, but now that's no longer the case.

Here it's a bit tricky, because StatsBase also gives you the various normalized / pdf / frequency options, smart bin selection, etc. I'm not 100% sure what is the best approach. I think Plots reimplemented a bunch of stuff (like a different bin selection algorithm, equal area histograms, etc.).

jkrumbiegel:
I think it’s a good idea for StatsMakie’s plotting functions to be absorbed by something else, be it a recipe package or AbstractPlotting. In light of the long compile times I was hoping we could add as many of the basic features to AbstractPlotting, to spare new users the hunt for even more packages to install. (You already have to import CairoMakie, AbstractPlotting, and MakieLayout for the basic subplots). Maybe I’m jaded because I have to recompile everything so often during development. Might be quite different for normal users.

By the way, how is AlgebraOfGraphics coming along? I saw there was not much happening in the repo in the last weeks, although the approach looks so promising :)

mkborregaard:
Plots used to have a StatsBase-free histogram function, but I talked Tom into taking on the dep - StatsBase is so basal as to be essentially a stdlib and it really makes sense to consolidate the histogram method across julia. Also because e.g. autobinning and normalizations are actually non-trivial (check eg the discussion on this PR https://github.com/JuliaStats/StatsBase.jl/pull/243 and this https://github.com/JuliaPlots/Plots.jl/pull/787 and this https://github.com/JuliaPlots/Plots.jl/pull/787 and https://github.com/JuliaPlots/Plots.jl/pull/713

Please let's not go down that entire road again by starting with some quick-and-dirty histogram implementation to get rid of a harmless dep.

asinghvi17:
I think that just slimming StatsMakie down by removing the grammar-of-graphics stuff would alleviate the load-time issue.  (Also, we should think about what inputs to allow - should `histogram` be able to do both 1D and 2D, or should there be a `histogram2d` recipe for that?  What about 3D (volume rendering)? 

@piever, it seems to me that AOG has feature-parity with StatsMakie.  Should we essentially "replace" the current StatsMakie by AOG, and then add only stats-related recipes to StatsMakie?

piever:
> By the way, how is AlgebraOfGraphics coming along? I saw there was not much happening in the repo in the last weeks, although the approach looks so promising :)

I'm finally happy with the design, but at the same time I got really busy with thesis writing, so I probably won't have time to dive back in for a a while. OTOH, I was planning to showcase it at juliacon, so it'll probably see a boost by then :)

> Also, we should think about what inputs to allow - should histogram be able to do both 1D and 2D, or should there be a histogram2d recipe for that? What about 3D (volume rendering)?

StatsMakie currently supports 1d, 2d, and 3d histograms with the same syntax, and I think it makes sense: the dimensionality is clear from how many inputs the user passes. Same for `density`: it should also accept 1d, 2d, and 3d inputs.

> @piever, it seems to me that AOG has feature-parity with StatsMakie. Should we essentially "replace" the current StatsMakie by AOG, and then add only stats-related recipes to StatsMakie?

I would think so, esp. since the implementation in StatsMakie is not as clean as the one in AlgebraOfGraphics. My current two concerns are:

1. would people find AlgebraOfGraphics too verbose? I suspect it's not as bad if we use `Scatter` as default plot type in AbstractPlotting, so then `scatter(Data(iris), Group(:Species), :SepalLength, :SepalWidth)` would become `data(iris) * style(:SepalLength, :SepalWidth, color = :Species => categorical)`

2. where would statistical analyses live?

By analyses I mean things like `linear` or `smooth`, where you are really transforming the data and returning a plottable object. Somehow it seems like it is a better fit to have them in AlgebraOfGraphcs, but then maybe it's a bit too much to force the user to load AlgebraOfGraphics for a linear regression. OTOH, having them there allows for nice, automatic legend entries and things like that: http://juliaplots.org/AlgebraOfGraphics.jl/dev/generated/tutorial/#Layout-1

SimonDanisch:
Great :) Can you add a couple of examples to `test/simple.jl`?

asinghvi17:
Well, this is awkward - the `histogram` export clashes with StatsMakie's histogram analysis.  Personally, I would rather refactor the algebra-of-graphics stuff out of StatsMakie, and just put statistical recipes there (similarly to StatsPlots.

jkrumbiegel:
It's a bit weird. On the one hand I'm all for bundling functionality in packages where it makes sense and relying on other people's packages for statistical analyses etc. On the other hand the compilation latency incurred by StatsMakie's dependencies is pretty massive. The most common histogram for example is the standard 1d histogram, by far. Now, do you want to load Optim.jl and Distributions.jl just to have complex but rare cases covered, or do you tier the methods by usage frequency and try to keep the core set lean? Given that ways of substantially reducing compile time are still further out, I think the most effective approach for now is to avoid depending on large packages wherever possible. And it was just my personal feeling that a bread-and-butter histogram should be one of the quickly loaded default methods, I don't want to load all of StatsMakie for that.

PR #440, opened 2020-05-29
Make the default marker size actually visible

asinghvi17:
0.1px is just...no...


SimonDanisch:
We don't have pixel space as default yet

PR #442, opened 2020-06-01
WIP: Adds attribute for setting width of 3D axis lines

grero:
This was partially prompted by a slight confusion in the documentation for Axis3d.frames.linewdith. In the docs, it says that this property controls the width of the axis lines, but in the code it actually sets the width of the gridlines instead. This is my attempt at clearing up that confusion by introducing the property `axislinewidth`, which does what the documentation claims that`linewidth` is supposed to do. 
I have yet to update the actual documentation, since I'm actually not sure how to do that. Of course, any comments and corrections are welcome!


SimonDanisch:
Thanks! :)

PR #443, opened 2020-06-01
Remove global theme support

SebastianM-C:
This breaks on julia 1.5 due to the `include` in `__init__`.
It may be simpler to explicitly have the user do the `include`. See also the corresponding MakieGallery PR


SimonDanisch:
Can we replace this with a warning if a config exists?

SebastianM-C:
Added the warning.

SimonDanisch:
Thank you :)

PR #449, opened 2020-06-04
Merge Makielayout with git history

SimonDanisch:


SimonDanisch:
Todos:

- [ ] include MakieLayout tests
- [ ] improve AbstractPlotting test suite while at it (let tests live here, only use MakieGallery for running them)
- [ ] update license file
- [ ] merge documentation
- [ ] assure that @jkrumbiegel authorship isn't lost

jkrumbiegel:
>  include MakieLayout tests
MakieLayout doesn't really have tests, the only tests check that the constructors of the different layoutables don't outright fail when called. That at least ensures that the observable pipelines inside didn't get messed up, so I get quite a lot of code coverage just by those couple of lines. I never had time to implement visual checks.

asinghvi17:
Nice!  We should make an announcement in the slack channel!

SimonDanisch:
Well, I think we should clean up the tests a bit first and then tag ;)

PR #450, opened 2020-06-09
add more tests

SimonDanisch:


test/runtests.jl
```diff
@@ -3,6 +3,11 @@ using MakieGallery
 using Test
 using GLMakie
 using StaticArrays, GeometryBasics
+using Observables
+using GeometryBasics: Pyramid
+using PlotUtils
+using MeshIO, FileIO
+using AbstractPlotting: Pixel
```
asinghvi17, line 10: do we not export Pixel?

test/runtests.jl
```diff
@@ -3,6 +3,11 @@ using MakieGallery
 using Test
 using GLMakie
 using StaticArrays, GeometryBasics
+using Observables
```
asinghvi17, line 6: We should reexport all these utility packages (Colors, Observables)

test/reference_image_tests/simple.jl
```diff
@@ -0,0 +1,208 @@
+using MakieGallery: @substep
+
+@block SimonDanisch ["tests"] begin
+    ## Scatter
+    @cell begin
```
asinghvi17, line 5: would multiple cells be better here?  all of these in one cell makes it kind of hard to debug.

asinghvi17: Also, we do need to separate 2d examples from 3d ones (for CairoMakie)

test/runtests.jl
jkrumbiegel: yes, `onany` and `Gray` are two missing exports that always annoy me :)

test/reference_image_tests/simple.jl
SimonDanisch: Yeah I need to figure out how to do this... I had to put them in one cell, to have theglobally shared functions and data work!

asinghvi17: You could also put toplevel code in, that works now

PR #457, opened 2020-06-11
[WIP] Patterns/Hatching

ffreyer:
@theogf asked about patterns/hatching on Slack/Zulip, stuff like https://tex.stackexchange.com/questions/183889/how-to-draw-dashed-area-in-graph-tikz

This is something that could be done using repeated textures. But afaik there is no way to tell GLMakie to repeat a texture. There's also no way to change other texture options, such as minimization or magnification methods. So I thought it'd be a good idea to add a Texture type to AbstractPlotting which can hold on to these options and forward them to `GLAbstraction.Texture` later down the line.

Here's an example of how you could generate a dashed pattern with this. (Though `uv_scale` should be something like `uv_scale = pixelwidths(filled_area) ./ size(img)` for correct scaling and the image could use some better anti-aliasing)

```julia
function smooth(A)
    N, M = size(A)
    out = similar(A, Float64)
    for i in 1:N, j in 1:M
        out[i, j] = (
            A[i, j] +
            A[mod1(i+1, N), j] +
            A[mod1(i-1, N), j] +
            A[i, mod1(j+1, N)] +
            A[i, mod1(j-1, N)]
        ) / 5
    end
    out
end
A = [mod1(x+y, 20) > 17 ? 0 : 1 for x in 1:20, y in 1:20]
img = map(x -> RGBA(0, 0, 0, 1-x), smooth(A))
scene = surface(1:10, 1:10, [0. for _ in 1:10, _ in 1:10], color=Texture(img, x_repeat=:repeat), shading=false, uv_scale=Vec2f0(50))
```

Screenshot from 2020-06-11 20-19-31

See also: https://github.com/JuliaPlots/GLMakie.jl/pull/108

### DIFF
--- a/src/Patterns.jl
+++ b/src/Patterns.jl
@@ -1,0 +1,80 @@
+# TODO move this?
+
+struct Pattern{T, D} <: AbstractArray{T, D}
+    img::Array{T, D}
+end
+
+function Pattern(
+        mask::Array{T, 2}, color1, color2; kwargs...
+    ) where {T <: AbstractFloat}
+    # to color
+    img = map(x -> to_color(color1) * x + to_color(color2) * (1-x), mask)
+    Pattern(img)
+end
+
+# Pattern from String/Character
+Pattern(style::String; kwargs...) = Pattern(style[1]; kwargs...)
+function Pattern(
+        style::Char = '/';
+        color1 = RGBA(0, 0, 0, 1), color2=RGBA(1, 1, 1, 0), kwargs...
+    )
+    mask = if style == '/'
+        LineMask(direction=Vec2f0(1); kwargs...)
+    elseif style == '\\'
+        LineMask(direction=Vec2f0(1, -1); kwargs...)
+    elseif style == '-'
+        LineMask(direction=Vec2f0(1, 0); kwargs...)
+    elseif style == '|'
+        LineMask(direction=Vec2f0(0, 1); kwargs...)
+    elseif style == 'x'
+        mask1 = LineMask(direction=Vec2f0(1); kwargs...)
+        mask2 = LineMask(direction=Vec2f0(1, -1); kwargs...)
+        map((x, y) -> clamp(x+y, 0, 1), mask1, mask2)
+    elseif style == '+'
+        mask1 = LineMask(direction=Vec2f0(1, 0); kwargs...)
+        mask2 = LineMask(direction=Vec2f0(0, 1); kwargs...)
+        map((x, y) -> clamp(x+y, 0, 1), mask1, mask2)
+    else
+        LineMask(; kwargs...)
+    end
+    Pattern(mask, color1, color2; kwargs...)
+end
+
+# Pattern with line in some direction
+# the line should be periodic in within a box of size tilesize
+function LineMask(;
+        direction = Vec2f0(1), width = 2f0, tilesize = (10,10),
+        shift=Vec2f0(0), kwargs...
+    ) where {CT <: Colorant}
+
+    mask = zeros(tilesize...)
+    width -= 1 # take away center
+    if abs(direction[1]) < abs(direction[2])
+        # m = dx / dy; x = m * (y-1) + 1
+        m = direction[1] / direction[2]
+        for y in 1:tilesize[2]
+            cx = m * (y-shift[2]) + shift[1]
+            r = floor(Int64, cx-0.5width):ceil(Int64, cx+0.5width)
+            for x in r[2:end-1]
+                mask[mod1(x, tilesize[1]), y] = 1.0
+            end
+            mask[mod1(r[1], tilesize[1]), y] = 1 - abs(cx-0.5width - r[1])
+            mask[mod1(r[end], tilesize[1]), y] = 1 - abs(cx+0.5width - r[end])
+        end
+    else
+        # m = dy / dx; y = m * (x-1) + 1
+        m = direction[2] / direction[1]
+        for x in 1:tilesize[1]
+            cy = m * (x-shift[1]) + shift[2]
+            r = floor(Int64, cy-0.5width):ceil(Int64, cy+0.5width)
+            for y in r[2:end-1]
+                mask[x, mod1(y, tilesize[2])] = 1.0
+            end
+            mask[x, mod1(r[1], tilesize[2])] = 1 - abs(cy-0.5width - r[1])
+            mask[x, mod1(r[end], tilesize[2])] = 1 - abs(cy+0.5width - r[end])
+        end
+    end
+
+    # TODO better blurring
+    mask
+end
--- a/src/Patterns.jl
+++ b/src/Patterns.jl
@@ -6,7 +6,7 @@ end
 
 function Pattern(
         mask::Array{T, 2}, color1, color2; kwargs...
-    ) where {T <: AbstractFloat}
+    ) where {T <: Real}
     # to color
     img = map(x -> to_color(color1) * x + to_color(color2) * (1-x), mask)
     Pattern(img)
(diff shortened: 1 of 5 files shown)


SimonDanisch:
Oh, we have that already... It's in ShaderAbstraction (I know, not the best place....):
https://github.com/SimonDanisch/ShaderAbstractions.jl/blob/master/src/types.jl#L67
That's the type that should be used to pass things through to opengl!

SimonDanisch:
But, it isn't hooked up 100% yet ... But it should already be hooked up in certain places 😅

ffreyer:
Oh, guess I can pivot to hooking that up then

ffreyer:
Well, since `ShaderAbstractions.Sampler <: AbstractArray`[convert_attributes](https://github.com/JuliaPlots/AbstractPlotting.jl/blob/0de466548de14719392595a1ba528d9f5f385890/src/conversions.jl#L592) gets applied and it loses all the Texture-relevant information...

asinghvi17:
You could specialize convert_arguments for textures?

asinghvi17:
also, how should we implement this for CairoMakie?

ffreyer:
I think all the repeated texture stuff  can move to GLMakie. Would it be easier to just support hatching in CairoMakie?

SimonDanisch:
If I remember correctly, most texture attributes should be supported by cairo pattern

ffreyer:
I implemented a `Pattern` struct which can be passed in `mesh` and `surface`.

Examples:
```julia
scene = surface(1:10, 1:10, [0 for i in 1:10, j in 1:10], shading=false, color=Pattern("+", shift=Vec2f0(1, 0)))
```
Screenshot from 2020-06-13 19-48-40

```julia
catmesh = FileIO.load(Makie.GLMakie.assetpath("cat.obj"))
scene = mesh(catmesh, color=AbstractPlotting.Pattern("x", color1 = :red, color2 = :yellow))
```
Screenshot from 2020-06-13 19-48-51

src/Patterns.jl
```diff
@@ -0,0 +1,80 @@
+# TODO move this?
+
+struct Pattern{T, D} <: AbstractArray{T, D}
+    img::Array{T, D}
+end
```
asinghvi17, line 5: We should also store the pattern type, for vector backends.

ffreyer: I don't know what vector backends would need. I guess for lines it would be the direction, linewidth, spacing and colors?

What about cross hatches? And what user defined patterns?

ffreyer: Maybe there should be a `LinePattern` and an `ImagePattern`

asinghvi17: Yeah, that would be pretty good, I think.  I don't think vector backends would need more than that.

asinghvi17:
Oh, by the way...how do you see this working with `transform_func`?  I guess a texture is in screen space, so it should look the same, but the area should change according to the transform?

ffreyer:
In matplotlib the hatching is the same with log and normal axis. Only the area changes. This should also be how it currently works here.

There are two new backend attributes that can affect this - `uv_scale` and `fetch_pixel`. The former is a multiplier to texture coordinates, so if `uv_scale = Vec2f0(2)` the pattern becomes half as large. `fetch_pixel` controls whether the texture is accessed with screen (pixel) coordinates or with the normal texture coordinates. If you set `fetch_pixel = false` (and `uv_scalle` to a large enough value) the pattern will do whatever the mesh/surface dictates:

```julia
scene = mesh(Sphere(Point3f0(0), 1f0), color=Pattern("-", color1=:red, color2=:yellow))
scene2 = mesh(Sphere(Point3f0(0), 1f0), color=Pattern("-", color1=:red, color2=:yellow), fetch_pixel=false, uv_scale=Vec2f0(50))
vbox(scene, scene2)
```
Screenshot from 2020-06-14 13-29-24

So if you use a mesh/surface with enough vertices (and `transform_func` applies to those) you can set `fetch_pixel = false` and get a transformed pattern.

src/Patterns.jl
```diff
@@ -0,0 +1,116 @@
...
+    widths = width isa Vector ? width : [width for _ in 1:N]
+    shifts = shift isa Vector{<:Vec2} ? shift : [shift for _ in 1:N]
+    colors = (to_color(color1), to_color(color2))
+
+    LinePattern{typeof(colors[1]), 2}(dirs, widths, shifts, tilesize, colors)
+end
+
+# Pattern from String/Character
+Pattern(style::String; kwargs...) = Pattern(style[1]; kwargs...)
+function Pattern(style::Char = '/'; kwargs...)
```
asinghvi17, line 60:
```suggestion
"""
    Pattern(style::Char = '/'; kwargs...)

Creates a line pattern based on the given argument.  
Available patterns are `'/'`, `'\\'`, `'-'`, `'|'`, `'x'`, and `'+'`.

All keyword arguments correspond to the keyword arguments for [`LinePattern`](@ref).
"""
function Pattern(style::Char = '/'; kwargs...)

File: src/Patterns.jl [hunk 3]
```diff
@@ -0,0 +1,116 @@
...
+function Pattern(mask::Array{<: Real, 2}; color1=RGBA(0,0,0,1), color2=RGBA(1,1,1,0))
+    img = map(x -> to_color(color1) * x + to_color(color2) * (1-x), mask)
+    ImagePattern(img)
+end
+
+image(p::ImagePattern) = p.img
+
+
+
+struct LinePattern{T, D} <: AbstractPattern{T, D}
```
asinghvi17, line 31: This could probably also use some docs.

src/Patterns.jl
```diff
@@ -0,0 +1,116 @@
+# TODO move this?
+
+abstract type AbstractPattern{T, D} <: AbstractArray{T, D} end
```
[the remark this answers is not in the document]
asinghvi17: Also, do we need the `D` here if we know that the pattern is always 2D?  Another option would be to say `AbstractPattern{T} <: AbstractArray{T, 2}`...

src/Patterns.jl
```diff
@@ -0,0 +1,116 @@
...
+end
+
+Pattern(img::Array{<: Colorant, 2}) = ImagePattern(img)
+
+function Pattern(mask::Array{<: Real, 2}; color1=RGBA(0,0,0,1), color2=RGBA(1,1,1,0))
+    img = map(x -> to_color(color1) * x + to_color(color2) * (1-x), mask)
+    ImagePattern(img)
+end
+
+image(p::ImagePattern) = p.img
```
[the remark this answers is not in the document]
ffreyer: Maybe `to_image`?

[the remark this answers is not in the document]
asinghvi17: Sounds good!

ffreyer:
My `show` or `display` method breaks `?some_plot`...

src/Patterns.jl
```diff
@@ -0,0 +1,116 @@
...
+    img::Array{T, D}
+    ImagePattern(img::Array{T, 2}) where {T <: Colorant} = new{T, 2}(img)
+end
+
+Pattern(img::Array{<: Colorant, 2}) = ImagePattern(img)
+
+function Pattern(mask::Array{<: Real, 2}; color1=RGBA(0,0,0,1), color2=RGBA(1,1,1,0))
+    img = map(x -> to_color(color1) * x + to_color(color2) * (1-x), mask)
+    ImagePattern(img)
+end
```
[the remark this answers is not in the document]
ffreyer: Currently stuff is sorted by type. First `AbstractPattern` functions, then `ImagePattern`, then `LinePattern`. Should I sort it to be types first, then constructors, then other functions?

SimonDanisch:
Will merge in: https://github.com/JuliaPlots/AbstractPlotting.jl/pull/460

PR #458, opened 2020-06-11
better macro hygiene so the macro works in submodules

jkrumbiegel:


jkrumbiegel:
this fixes `@lift` uses in the AbstractPlotting.MakieLayout submodule

PR #464, opened 2020-07-03
add errorbar recipe

jkrumbiegel:
this is an errorbar recipe with a couple of input options and the ability to do horizontal or vertical bars. the whiskers are also given in screen space units, which is practical because whisker width has no meaning in terms of data scaling. the implementation might be debatable, and maybe this can be done with `to_world` or something like that, but I didn't get those functions to do what I wanted and they weren't documented.

we really need default functions to easily go back and forth between different scenes and the screen coordinates

### DIFF
--- a/src/basic_recipes/errorbars.jl
+++ b/src/basic_recipes/errorbars.jl
@@ -1,0 +1,99 @@
+@recipe(Errorbars) do scene
+    Theme(
+        whiskerwidth = 5,
+        color = :black,
+        linewidth = 1,
+        direction = :y
+    )
+end
+
+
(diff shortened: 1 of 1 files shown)


piever:
I guess this comes from the corresponding recipe in StatsMakie? In general, I agree  that those should live here. I've made a issue over at Makie to discuss the whole StatsMakie situation: https://github.com/JuliaPlots/Makie.jl/issues/636.

SimonDanisch:
Would be nice to have some basic tests for it, probably here: 
https://github.com/JuliaPlots/AbstractPlotting.jl/blob/master/test/reference_image_tests/simple.jl

jkrumbiegel:
No this one I made "from scratch" just because of the screen space whiskers. It's not complicated anyway but yeah would be nice not to depend on statsmakie for that

PR #467, opened 2020-07-07
better scatter defaults

jkrumbiegel:
This should provide better defaults for less tweaking with simple plots. The new defaults would have these advantages:

- no squished markers
- reasonable default size
- size not dependent on data scaling (you can't have a good default there, and 0.1 is bad for everything but our examples)
- stroke enabled by default for some robustness against overplotted markers
- gray65 fill for good contrast against light and dark backgrounds and a more elegant look than black

Here's a before just plotting 500 randn() points:

old

And after:

new

### DIFF
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -200,13 +200,13 @@ $(ATTRIBUTES)
 @recipe(Scatter, positions) do scene
     Attributes(;
         default_theme(scene)...,
-        color = :black,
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
so what is the process of merging this? do we generate all the new reference images manually and replace them in the test repo? maybe some tests have to be changed because they rely on the previous defaults

PR #471, opened 2020-07-10
Add visible attribute to errorbar

ffreyer:
I tried making a figure where you can toggle the visibility of various plots and noticed that `errorbars` don't support `visible` yet. So here's a pr to fix that. Though I wonder if it would make more sense to have `default_theme(scene, LineSegments)` in there instead, similar to how it's done [here](https://github.com/JuliaPlots/AbstractPlotting.jl/blob/4d25c309320953076f78be74e64c70308933c6a3/src/basic_recipes/basic_recipes.jl#L452).

### DIFF
--- a/src/basic_recipes/errorbars.jl
+++ b/src/basic_recipes/errorbars.jl
@@ -3,7 +3,8 @@
         whiskerwidth = 10,
         color = :black,
         linewidth = 1,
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Thanks!

PR #472, opened 2020-07-11
add LEditBox

ffreyer:
I tried making an edit box. The basic functionality works - it displays text, you can select it by clicking on it, type and remove text and deselect it by pressing enter or clicking elsewhere. You can also pass a custom keyboard layout if you want.

Some issues this implementation has:
* Keyboard layouts are rather cumbersome and ignore the system language.
* The keyboard layout is probably incomplete
* You cannot mouse drag select text. 
* There is no cursor. (It can't be moved, it isn't displayed and sorta as a result you cannot use the delete key)
* No ctrl + a, c, v, etc
* No shift+alt modifier (though that'd be rather easy to add)
* Typing in the edit box doesn't "consume" events, so you may trigger other events in the scene while typing.
* Text overflows in both directions, rather than just towards the left. So if you type a lot you won't see the most recently typed characters
* If you press a second key before releasing the first you'll get ghost inputs. I.e. `a down, b down, a up, b up` generates "abab" and `a down, b down, b up, a up` generates "abaa" when both should be "ab".

For most of these issues I either don't know how to solve them or I think they're low priority. Maybe someone else can take a stab at it or we can leave them later?

### DIFF
--- a/src/makielayout/MakieLayout.jl
+++ b/src/makielayout/MakieLayout.jl
@@ -60,6 +60,7 @@ include("lobjects/llegend.jl")
 include("lobjects/lobject.jl")
 include("lobjects/lscene.jl")
 include("lobjects/lmenu.jl")
+include("lobjects/leditbox.jl")
 
 export LAxis
 export LSlider
@@ -72,6 +73,7 @@ export LLegend
 export LegendEntry, MarkerElement, PolyElement, LineElement, LegendElement
 export LScene
 export LMenu
+export LEditBox
 export linkxaxes!, linkyaxes!, linkaxes!
 export AxisAspect, DataAspect
 export autolimits!, limits!
--- a/src/makielayout/defaultattributes.jl
+++ b/src/makielayout/defaultattributes.jl
@@ -746,3 +746,121 @@ $(let
 end)
 """
 LScene
+
+
+DEFAULT_KEYBOARD_LAYOUT = let
+    # Key => (base, shift-mod, altgr-mod)
+    Key = AbstractPlotting.Keyboard
+
+    Dict(
+        Key.q => ('q', 'Q'),
+        Key.w => ('w', 'W'),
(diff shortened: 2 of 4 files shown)


SimonDanisch:
Nice!! :)

jkrumbiegel:
Oh I was already relatively far with my own implementation https://github.com/jkrumbiegel/MakieLayout.jl/blob/jk/ltextbox/src/lobjects/ltextbox.jl Did you see that or just straightaway implemented your own version?

I think you don't really need to deal with keyboard layouts, as there is a unicode symbol event that should receive the correct characters without you having to deal with correctly assigning them. I also had a cursor implemented, but it was still missing drag to select, as that's pretty complicated.

ffreyer:
> Oh I was already relatively far with my own implementation https://github.com/jkrumbiegel/MakieLayout.jl/blob/jk/ltextbox/src/lobjects/ltextbox.jl Did you see that or just straightaway implemented your own version?

No, I didn't see that. I had implemented something like this a long time ago, detached from MakieLayout, and tried to update it. 

> I think you don't really need to deal with keyboard layouts, as there is a unicode symbol event that should receive the correct characters without you having to deal with correctly assigning them. 

Oh wow, I somehow completely missed that.

Anyway, your implementation looks better so I'm closing this. I'd love to see it soon, even without drag-selection.

SimonDanisch:
@jkrumbiegel, should we try to have this here, even if its not 100% finished?

jkrumbiegel:
Sure, I've made a PR https://github.com/JuliaPlots/AbstractPlotting.jl/pull/474

> Anyway, your implementation looks better so I'm closing this. I'd love to see it soon, even without drag-selection.

Thanks for your effort! I just wanted to make sure you don't spend your time on this, too :) I think it's already quite useable, at least when I just tried it again it seemed ok.

PR #473, opened 2020-07-12
fix delete_scene

piever:
I think what's going on is that `scene.plots` is getting changed in place during the iteration, so the iterator will think it has finished when it has not. For example, if `length(scene.plots) == 2`, after cleaning up the first, it will see that the length of the array equals the number of times it has iterated, so it will stop. I imagine the shallow copy is not a performance problem, otherwise one can use some smart while loop I guess, but I'd prefer to keep it simple.

### DIFF
--- a/src/makielayout/helpers.jl
+++ b/src/makielayout/helpers.jl
@@ -338,7 +338,7 @@ function remove_element(::Nothing)
 end
 
 function delete_scene!(s::Scene)
-    for p in s.plots
+    for p in copy(s.plots)
         delete!(s, p)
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
ah that makes sense, that happened to me before :) although I'm wondering why this lives in makielayout? I don't remember under what circumstances I wrote this

piever:
Yes, now that the two packages are merged it should just be `Base.delete!(s::Scene)` and live outside MakieLayout, IMO.

PR #474, opened 2020-07-13
add LTextbox

jkrumbiegel:
I honestly forgot what the things were holding this up. I think one thing was that the cursor would be better placed at the right advance of the selected character, now the line is too far left if you type a space for example. Then I'm not sure if the animation loop of the cursor will go on forever if the scene is destroyed while the textbox is active. Other than that, the base functionality seems to work. Oh and what's a bit weird is the internal logic handling an empty text box, I have to cheat and insert a space when that happens because text objects error if they're empty.

### DIFF
--- a/src/makielayout/MakieLayout.jl
+++ b/src/makielayout/MakieLayout.jl
@@ -61,6 +61,7 @@ include("lobjects/llegend.jl")
 include("lobjects/lobject.jl")
 include("lobjects/lscene.jl")
 include("lobjects/lmenu.jl")
+include("lobjects/ltextbox.jl")
 
 export LAxis
 export LSlider
@@ -73,6 +74,7 @@ export LLegend
 export LegendEntry, MarkerElement, PolyElement, LineElement, LegendElement
 export LScene
 export LMenu
+export LTextbox
 export linkxaxes!, linkyaxes!, linkaxes!
 export AxisAspect, DataAspect
 export autolimits!, limits!
--- a/src/makielayout/MakieLayout.jl
+++ b/src/makielayout/MakieLayout.jl
@@ -14,6 +14,7 @@ import Animations
 import PlotUtils
 using GridLayoutBase
 import Showoff
+using Colors
 
 const FPS = Node(30)
 const COLOR_ACCENT = Ref(RGBf0(((79, 122, 214) ./ 255)...))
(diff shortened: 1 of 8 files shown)


ffreyer:
I don't think discarding the input when clicking away or pressing escape is a good (default) behavior. Most text fields I can think of keep their content in either case.

jkrumbiegel:
Hm maybe you're right, the motivation was that I thought you want to enter some text and then submit it to have an effect. Not really implement html forms where you might fill in this and return to that etc. So the textbox should stay in a valid state, either you change its content or you don't and then it jumps back to what it was. Otherwise you can get out of sync, what's displayed will not be what is "active". Currently, the main observable for dealing with the content is textbox.content and it is only updated when submitting with enter.

ffreyer:
Imo `content` should just be whatever is currently shown in the textbox. You could have a `content` and `confirmed_content` observable to allow both options. Or you could have some observable that triggers on `enter` so that people can implement a `confirmed_content` observable if they need it.

jkrumbiegel:
Could you describe a use case how you'd utilize an observable that always updates with every character change? The ones that I can come up with all align more with the "change and submit" approach

jkrumbiegel:
I also have a validator attribute where you can restrict the possible inputs, so you could for example only enter valid floats. That would also not work with a continuous update

ffreyer:
> Could you describe a use case how you'd utilize an observable that always updates with every character change? The ones that I can come up with all align more with the "change and submit" approach

Maybe auto-completion or searches? 

My concern is not so much what you can do with it though, but rather that it's annoying to use. I built some sort of GUI for evaluating simulations before, showing a bunch of line plots you can turn on and off with buttons, alongside two 3D plots. I had a bunch of text edit fields there, holding numbers and names for things. I can see myself writing some name in an edit field and clicking and dragging some 3D plot around, or turning some lines on or off to confirm what I've written, losing whatever I've written because I didn't press enter.

piever:
In Interact textboxes have two observables, one that updates with every key press (the value), and one that simply triggers on the HTML "change" event (which I think is when you focus out of the thing). Maybe we need two observables here as well?

I remember though that this approach required a bit of care with validators (say input `type=number`), but it's feasible.

jkrumbiegel:
Ok I think I got this to a pretty good state (minus the empty string workarounds, but ok)

I've incorporated the feedback about the text deletion on defocus. Now the behavior is like this:

There is an attribute `reset_on_defocus`, which when `true` means that exiting the textbox resets it to the current `stored_string`. That setting is `false` by default, meaning no text is deleted. Text in a defocused textbox is grayed out if it's the placeholder text or user defined text that doesn't match the `stored_string`, therefore indicating that it's not submitted, yet.

The changes in displayed texts can be listened to via the attribute `displayed_string` if that's needed.

PR #476, opened 2020-07-15
drop async from LSlider

ffreyer:
LSlider becomes very slow when triggering more expensive calculations. I tried using it to trigger a ~1.5ms calculation which and it took about 5s to update. This is fixed by removing the `@async`.

With `@async`: (Updates on mouse down are always fast)
LSlider_with_async

Without `@async`:
LSlider_no_async

### DIFF
--- a/src/makielayout/lobjects/lslider.jl
+++ b/src/makielayout/lobjects/lslider.jl
@@ -135,14 +135,12 @@ function LSlider(parent::Scene; bbox = nothing, kwargs...)
             dif[2] / (height(sliderbox[]) - 2pad)
         end
         if fraction != 0.0f0
-            @async begin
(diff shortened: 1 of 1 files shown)


jkrumbiegel:
oh interesting, I didn't see this PR until now. I actually put in the @async thing because it seemed to make it faster for me, but maybe it needs to be reverted. Or some changes in GLMakie happened in the meantime, there were of course multiple changes in the rendering pipeline

ffreyer:
Seems like this is related to vsync. It's smooth for me with or without this pr if vsync is off, but it slows down on master with `vsync=true`.

jkrumbiegel:
Interesting, it seems that glfw gsync blocks the whole thread and coroutines can't switch anymore. If the render thread was really a thread I guess this wouldn't be a problem

PR #478, opened 2020-07-22
port recipes from StatsMakie

piever:
This is part of the overall plan in https://github.com/JuliaPlots/Makie.jl/issues/636 to deprecate StatsMakie.

Status:

- [x] `boxplot` (requires StatsBase)
- [x] `crossbar` (requires StatsBase)
- [x] `violin` (requires KernelDensity)
- [x] `convert_arguments` for a `kernel density` object (requires KernelDensity)
- [x] distribution plots and `qqplot` (requires Distributions)
- [x] `histogram` (it's here already, but I would like to port the current recipe to use StatsBase and export it)
- ~support "dodge and stack" for `barplot` and histogram and "dodge" for `violin`, `boxplot`, and `crossbar`~

Clean up:

- [x] delete the ["multiple plot" machinery](https://github.com/JuliaPlots/AbstractPlotting.jl/blob/master/src/basic_recipes/multiple.jl), as it was only being used by StatsMakie, and it is a bit hacky.

Other things seem less relevant. All the "analyses" should, IMO, go in AlgebraOfGraphics.
As for the rest:

- `errorbar` was already ported here.
- `ribbon` and `shaded_background` seem less crucial (they are really just `lines + band`).

Overall the extra dependency load is quite small (StatsBase is pretty lightweight, Distributions is OK, KernelDensity will be OK after https://github.com/JuliaStats/KernelDensity.jl/pull/84).


piever:
I'm not sure how we want to go ahead with the merging. The main problem I guess is export conflicts with StatsMakie. I think the best is to simply bump the minor version, so that StatsMakie cannot get called with these exports.

piever:
In the end, I'm no longer sure whether the "dodge and stack" functionality belongs here or not. Curious to know what other people think, but I suspect this may go in without that. I can try and implement it from the AlgebraOfGraphics side and see how that goes.

src/stats/histogram.jl
```diff
@@ -0,0 +1,76 @@
...
+   norm 1.
+*  `:none`: Do not normalize.
+
+## Attributes
+$(ATTRIBUTES)
+"""
+@recipe(Histogram, values) do scene
+    Attributes(
+        bins = 15, # Int or iterable of edges
+        normalization = :none
```
piever, line 34: I've changed it from `relative = false` as it was before, because I've realized that StatsMakie and AbstractPlotting were defaulting to two different normalizations (StatsMakie was doing `pdf`, whereas here it was `probability`). I guess it's safer to be explicit because different people may have different expectations.

src/makielayout/lobjects/llegend.jl
```diff
@@ -358,7 +358,7 @@ function legendelements(plot::Scatter)
         strokecolor = plot.strokecolor)]
 end
 
-function legendelements(plot::Poly)
+function legendelements(plot::Union{Poly, Violin, BoxPlot, CrossBar})
```
piever, line 361: I've added `legendelements` for the new recipes. Should `Histogram` also have a `legendelements` method?

jkrumbiegel: maybe it would also be possible to define `legendelements` for all objects generically, where it tries to traverse all contained plots and layers them if they have `legendelements` defined? Because histograms are just barplots, maybe for simple versions that would work.
the problem is always when a barplot has multiple different colors, then there's no easy mapping

piever: Yes, that's a good point, I can wait to add the `legendelements` from `Histogram`, because it should somehow "follow" from the bar plot.

piever:
Actually, I wanted to ask: would it be possible to rename the `histogram / Histogram` recipe? Somehow I feel that it is a name that causes a few conflicts. On the one hand, given that we also export the plot type, it becomes clumsy to fit the histogram using StatsBase:

```julia
julia> using AbstractPlotting, StatsBase

julia> fit(Histogram, rand(100))
WARNING: both StatsBase and AbstractPlotting export "Histogram"; uses of it in module Main must be qualified
ERROR: UndefVarError: Histogram not defined
```

On the other hand, I would like to keep the lowercase name `histogram` for the analysis (just the data transformation, not the plotting) in AlgebraOfGraphics.

Possible proposals for the name:

- `hist / Hist` (as in matplotlib)
- `histogramplot / HistogramPlot`, so the distinction between making a plot (`HistogramPlot`) and just doing the statistics (`fit(Histogram)`) becomes clear.

I think the latter is more consistent with AbstractPlotting and, e.g., `BarPlot`, whereas I don't think we use abbreviations elsewhere.

jkrumbiegel:
hm `histogramplot` feels a bit long, maybe `histplot`? I honestly think we shouldn't even export all the plot type names, I've personally never needed them in user code. `Lines`, `Scatter`, `BarPlot`, etc. What do you think @SimonDanisch?

piever:
I've mentioned this on slack as well, but I was thinking that another option would also be to simply define a `convert_arguments` for `StatsBase.Histogram` (see [here](https://github.com/JuliaPlots/AbstractPlotting.jl/pull/478/files#diff-412054d0c5a39e802fbeab48e4ff3c61R3)), and then just do `plot(fit(Histogram, x))`, maybe also allowing `plot(Histogram, x)`. 

This way, we avoid adding a recipe that has very little to do with plotting, this feels more like a `convert_arguments` thing. Especially because for `2D` histograms, there will be a bit of ambiguity as to what plotting function to use (`heatmap`, `surface`, `contour`, etc...).

piever:
`histogram / Histogram` are unexported, so that they don't conflict for now. I am personally in favor of removing the recipe altogether, now that `StatsBase.Histogram` is plottable (in 1, 2, or 3D).

Regarding `dodge / stack`, they are not being ported, but I believe that they do not belong here. `dodge` is implemented in AlgebraOfGraphics and I suspect `stack` can also be added there.

I have rebased and think this is good to go.

cc: @SimonDanisch

piever:
A newer KernelDensity has been released (without the Optim dependency), so the load time issue should be mitigated (see https://github.com/JuliaStats/KernelDensity.jl/pull/84#issuecomment-669882821 for some timings).

I'm not too sure what's happening with the gitlab pipeline, cc @asinghvi17

PR #479, opened 2020-07-23
Overload convert_arguments for LineString

Sov-trotter:
This is the very first attempt at supporting GeometryBasics geometries based on https://github.com/JuliaPlots/Makie.jl/issues/660. Is there any better method to get to points? A set of decompose/reinterpret methods?


src/conversions.jl
```diff
@@ -284,6 +284,13 @@ function el32convert(x::AbstractArray{T, N}) where {T<:Union{Missing, <: Number}
     end::Array{Float32, N}
 end
 
+"""
+Takes an input `LineString` and decomposes it to points.
+"""
+function convert_arguments(PB::PointBased, linestring::LineString)
+    return convert_arguments(PB, linestring.points.parent.data)
```
SimonDanisch, line 291: How about using `decompose(Point, linestring)` ?

Sov-trotter: So that gives an error, 
```julia
ls1=LineString([Point(1, 2), Point(4,5), Point(10, 8), Point(1, 2)])

decompose(Point, decompose(GB.Ngon, ls1))
ERROR: MethodError: no method matching coordinates(::Array{GeometryBasics.Ngon,1})
Closest candidates are:
  coordinates(::GeometryBasics.Ngon) at /home/ezio/.julia/packages/GeometryBasics/CMxB4/src/basic_types.jl:76
  coordinates(::Simplex) at /home/ezio/.julia/packages/GeometryBasics/CMxB4/src/basic_types.jl:146
  coordinates(::LineString) at /home/ezio/.julia/packages/GeometryBasics/CMxB4/src/basic_types.jl:184
  ...
Stacktrace:
 [1] decompose(::Type{Point}, ::Array{GeometryBasics.Ngon,1}) at /home/ezio/.julia/packages/GeometryBasics/CMxB4/src/interfaces.jl:104
 [2] top-level scope at REPL[59]:1
```

So this is quite synonymous to https://github.com/JuliaGeometry/GeometryBasics.jl/issues/69

Sov-trotter: Also the `decompose` method seems to be slow, and in our case we will need some nested decomposition maybe.

Sov-trotter:
Also I was wondering, what is the correct way to support a vector of defined geometries. One way is iterating manually, but how can that be achieved via `convert_arguements`. I tried writing one for the above `LineString` method but that only returned the first `LineString`.

src/conversions.jl
Sov-trotter: https://github.com/JuliaGeometry/GeometryBasics.jl/pull/73

src/conversions.jl
```diff
@@ -284,6 +284,13 @@ function el32convert(x::AbstractArray{T, N}) where {T<:Union{Missing, <: Number}
     end::Array{Float32, N}
 end
 
+"""
+Takes an input `LineString` and decomposes it to points.
+"""
+function convert_arguments(PB::PointBased, linestring::LineString)
```
asinghvi17, line 290: I think you should just define a convert for a vector of linestrings as well.  Would that be a multilinestring or something else?

Sov-trotter: Yeah. It would facilitate a `MultiLineString`. But it's tricky to get it to work. Since `convert_arguments` is called only once what is the right way of telling `AbstractPlotting` that one `LineString` ends here and the next starts here. The docs specified that adding `NaN` is a way, but the syntax is a bit unclear.

asinghvi17: You can just separate the lines using `Point2f0(NaN)`, I think

Sov-trotter:
@asinghvi17  can you take a look?

Sov-trotter:
decompose(PB, LineString) depends on https://github.com/JuliaGeometry/GeometryBasics.jl/pull/73

Sov-trotter:
What changes might go in here, as we are moving towards https://github.com/JuliaGeometry/GeometryBasics.jl/pull/80 in GeometryBasics?
Will the decompose method do the work for us?

SimonDanisch:
continued in: https://github.com/JuliaPlots/AbstractPlotting.jl/pull/489

PR #482, opened 2020-08-02
fix mousestatemachine

jkrumbiegel:
Problems fixed:

- mousedown would trigger twice when starting a drag, causing interaction bugs
- mouse movements could trigger twice, causing performance drops
- first drag samples could be dropped, causing interaction bugs

I saw one weird problem with sliders in a very complex example that wouldn't drag on this version, but I could not reproduce it with normal examples, sliders work fine there. So it is possible that it's not related to this implementation at all but because of some other bug in that complex example. As I can't reproduce it, I think this should be fine


jkrumbiegel:
seems to fix https://github.com/jkrumbiegel/MakieLayout.jl/issues/73

PR #486, opened 2020-08-09
Support GeometryBasics Polygon

Sov-trotter:
So this overload convert_arguments for a polygon/multipolygon. (depends on #479 )


Sov-trotter:
As we are moving towards https://github.com/JuliaGeometry/GeometryBasics.jl/pull/80 in GeometryBasics?
Is the current method good?

SimonDanisch:
continued in: https://github.com/JuliaPlots/AbstractPlotting.jl/pull/489

PR #487, opened 2020-08-11
make scatter the default plot

piever:
I think `lines` is not such a great default for arbitrary vectors of numbers, because they are probably not sorted, so one just gets a garbled thing. Plus, there's no reason to restrict ourselves to numbers now that we support categorical data.

### DIFF
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -472,8 +472,7 @@ end
 """
     `plottype(plot_args...)`
 
-The default plot type for any argument is `lines`.
-Any custom argument combination that has only one meaningful way to be plotted should overload this.
+Any custom argument combination that has a preferred way to be plotted should overload this.
 e.g.:
 ```example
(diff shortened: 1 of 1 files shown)


src/interfaces.jl
```diff
@@ -472,8 +472,7 @@ end
 """
     `plottype(plot_args...)`
 
-The default plot type for any argument is `lines`.
```
piever, line 475: I don't really understand where this is defined. I though the default was to error.

src/interfaces.jl
```diff
@@ -472,8 +472,7 @@ end
...
-Any custom argument combination that has only one meaningful way to be plotted should overload this.
```
piever, line 476: I changed the wording. IMO, as long as one method is _preferred_, one should overload this. The user can force their preference regardless, so we don't lose anything by having a sensible default.

SimonDanisch:
Thanks :)

PR #490, opened 2020-08-19
Removed strokes from arrowheads

vilim:
Fixes the issue where arrowheads had strokes, causing rendering errors (a regression introduced by changing the default scatter style)


SimonDanisch (approved):

SimonDanisch:
Thanks!

PR #491, opened 2020-08-25
Don't export `transform!` (conflicts with DataFrames.jl)

greimel:
fixes https://github.com/JuliaPlots/Makie.jl/issues/589

this was discussed on Slack a while ago. It is slightly annoying because `transform!` is a central function for working with DataFrames.


jkrumbiegel:
I support this. @SimonDanisch do you think it's important to keep? I have never used it.

SimonDanisch:
Sounds fair to me!

PR #494, opened 2020-08-29
add hvlines for LAxes

jkrumbiegel:


src/makielayout/helpers.jl
```diff
@@ -433,3 +433,54 @@ function labelslider!(scene, label, range; format = string,
     layout = hbox!(label, slider, valuelabel; layoutkw...)
     (slider = slider, label = label, valuelabel = valuelabel, layout = layout)
 end
+
+
+
+
+function _hvlines!(ax::LAxis, ::Val{DIR}, datavals, axmins, axmaxs; attributes...) where DIR
```
SimonDanisch, line 440: I don't really like `_` for function prefixes like this...especially since this seems like a prefectly fine highlevel function?
Also, I probably wouldn't use `Val` for these kind of things anymore. There should be almost no performance penalty to have this as a normal variable, but there will be quite a bit of compilation penalty ;)

jkrumbiegel: Hm yeah could be that the 1 or 2 would even be constant folded as Int arguments, I can change it.

About the underscores, I thought that just communicates that this is a helper function for other named functions? I don't really mind either way but I think we can do better with clarifying what our internal functions are for. I get quite confused reading the source sometimes what's meant for what purpose

SimonDanisch (dismissed):

SimonDanisch (approved):

PR #497, opened 2020-09-14
move docs from makiegallery to abstractplotting

jkrumbiegel:


SimonDanisch:
Yay :)

SimonDanisch (approved):

MichaelHatherly:
Hey @jkrumbiegel @SimonDanisch, I came across your `Gitlab <: Documenter.DeployConfig` implementation when trying to solve the same problem with the Pumas docs (since they also get built on gitlab and hosted on github). Would you be interested in upstreaming this into Documenter rather than us having to re-implement it multiple times? If you've not got the time to PR I'm happy to do the work upstreaming it myself.

jkrumbiegel:
Would be cool if Documenter had this functionality, I was just not sure if it's "mainstream" enough of an approach? That's why I didn't make a PR, but you're welcome to do that!

MichaelHatherly:
I'll make a PR then, and we can see if two is mainstream enough.

PR #500, opened 2020-09-17
add contourf plot using isoband

jkrumbiegel:


SimonDanisch:
Should we merge this?

jkrumbiegel:
it basically works, but you can't change the data so that there's a different number of polygons, because then there are backend errors. would be great if you took a look at why that is and how it can be fixed (I guess it's the basic problem that number of polys and number of colors go out of sync somewhere)

PR #506, opened 2020-09-24
Fix breakage in update_limits!

timholy:
Oops...not off to a good start, as #504 looks like it broke stuff.

This time I added some tests to make sure the refactoring didn't break anything.


SimonDanisch:
Hm! I'd thought our simple tests would at least catch a simple refactor like that -.- 
Thanks for the fix & tests! :)

PR #513, opened 2020-09-26
Axis3D fixes/adjustments

ffreyer:
Sometimes axis grid lines end up outside the region suggested by the main axis lines. I fixed that by setting the origin to the minimum of the tick positions and data limits. I also adjusted the default spacing of tick labels and axis names to bit a larger and removed the outer most grid line. 

Before:
Screenshot from 2020-09-26 17-49-13

After:
Screenshot from 2020-09-26 17-44-19

### DIFF
--- a/src/basic_recipes/axis.jl
+++ b/src/basic_recipes/axis.jl
@@ -133,7 +133,7 @@ $(ATTRIBUTES)
             textsize = (6.0, 6.0, 6.0),
             align = axisnames_align3d,
             font = lift(dim3, theme(scene, :font)),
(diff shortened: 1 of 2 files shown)


ffreyer:
I'll try to fix the test failures...

ffreyer:
The test failure was about tick labels and positions of `scene = scatter(1:10, 1:10, 1:10)`. I adjusted the test because the ticks in the test do not match what actually gets displayed in the scene.

SimonDanisch (approved):

PR #515, opened 2020-09-28
fix crash on data update in errorbars

ffreyer:
The following crashes for me:

```julia
ps = Node(rand(Point2f0, 10))
errs = Node(rand(10))
errorbars(ps, errs)
ps.val = rand(Point2f0, 5)
errs.val = rand(5)
ps[] = ps[]
```

To fix this I'm restricting the number of `linesegpairs` generated to the minimum number of points available.


jkrumbiegel:
thanks for the heads up, I fixed that issue more "at the root" here https://github.com/JuliaPlots/AbstractPlotting.jl/pull/516 so I'll close this PR :)

PR #517, opened 2020-10-01
Addition of additive blending to volume rendering

vilim:
Additive blending, good for sparse, bright, volumes on dark backgrounds. Goes together with  JuliaPlots/GLMakie.jl#129


SimonDanisch:
Cool, thank you :)

PR #520, opened 2020-10-12
cleanups and inference improvements in makielayout

jkrumbiegel:


timholy (dismissed):

timholy:
Thanks! Just for interest, https://github.com/JuliaLang/julia/pull/38080 will allow decoupling constant prop from inlining. But this implementation seems like the right one for now.

PR #522, opened 2020-10-16
Add some precompiles relevant for layouts

timholy:
This shaves about 3s off `LAxis(scene, title = " ")`


SimonDanisch:
Thank you :)

PR #524, opened 2020-10-17
Fix and test pan/zoom

timholy:
This turns on selection-zoom (rubber band selection) for LAxis. It also ensures that if you zoom in and then zoom out you get back to where you started. Formerly we got to 0.9*1.1 == 0.99, this changes it to 0.9 * (1/0.9) == 1.0.

Also, formerly the rubber band was not being drawn. I'm guessing there was an API change so that `lines!(ax, ::Rect)` no longer traces the boundary of the rectangle. I changed it to `poly!`, but let me know if there's a different preferred strategy.

Fixes https://github.com/JuliaPlots/Makie.jl/issues/728

### DIFF
--- a/src/interaction/interactive_api.jl
+++ b/src/interaction/interactive_api.jl
@@ -146,15 +146,15 @@ rectangle has area > 0.
 
 The `kwargs...` are propagated into `lines!` which plots the selected rectangle.
 """
-function select_rectangle(scene; kwargs...)
+function select_rectangle(scene; strokewidth = 3.0, kwargs...)
     key = Mouse.left
     waspressed = Node(false)
     rect = Node(FRect(0, 0, 1, 1)) # plotted rectangle
     rect_ret = Node(FRect(0, 0, 1, 1)) # returned rectangle
 
     # Create an initially hidden rectangle
-    plotted_rect = lines!(
-        scene, rect, raw = true, visible = false, color = RGBAf0(0.1, 0.1, 0.8, 0.5), kwargs...,
+    plotted_rect = poly!(
+        scene, rect, raw = true, visible = false, color = RGBAf0(0, 0, 0, 0), strokecolor = RGBAf0(0.1, 0.1, 0.8, 0.5), strokewidth = strokewidth, kwargs...,
     )[end] # Why do I have to do [end] ?
 
     on(events(scene).mousedrag) do drag
--- a/src/makielayout/lobjects/laxis.jl
+++ b/src/makielayout/lobjects/laxis.jl
@@ -438,7 +438,7 @@ end
 
 Set the target limits of `la` to an automatically determined rectangle, that depends
 on the data limits of all plot objects in the axis, as well as the autolimit margins
-for x and y axis. 
+for x and y axis.
 """
 function autolimits!(la::LAxis)
 
@@ -681,7 +681,10 @@ function add_zoom!(ax::LAxis)
             pa = pixelarea(scene)[]
 
             # don't let z go negative
-            z = max(0.1f0, 1f0 + (zoom * zoomspeed))
+            z = max(0.1f0, 1f0 - (abs(zoom) * zoomspeed))
+            if zoom > 0
+                z = 1/z   # sets the old to be a fraction of the new. This ensures zoom in & then out returns to original position.
+            end
 
             mp_axscene = Vec4f0((e.mouseposition[] .- pa.origin)..., 0, 1)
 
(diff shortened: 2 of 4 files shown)


SimonDanisch:
Nice PR, thanks a lot! :)

SimonDanisch:
Ah the `lines!` change might actually be a bug, which should rather be fixed correctly, and meanwhile worked around by using `linesegments!`, which should be the better function here anyways!

jkrumbiegel:
In principle this is cool, and thanks for the work! But one issue I have with making leftclick-dragging have a default action like this, is that it will trigger also if a user puts a similar action onto a plot object (for example to drag elements around like in InteractiveChaos). So this would break such plots right now. We don't have a way to differentiate between scene interaction and plot-object-in-scene interaction, or rather dragging the scene is also draggin plot objects in the scene, but not the other way around.. Browsers have a different model with their event bubbling.

I did not enable this behavior also because I was not clear yet on a way to add and remove it as you go along. Right now, there's no way to remove zoom behavior if a user doesn't want it. I would actually prefer a range of options that can be turned on and off, like in other frameworks where a small gui let's you select interaction options.

One more thing that I don't really like is that this adds a rectangle to the scene all the time, I'd like it if on interaction the rectangle got added and then removed when the interaction is done. In general, in Makie we haven't paid enough attention to cleanup and these things accumulate :)

timholy:
I wondered about the left-click interaction myself, but since it's a documented Makie feature I felt it would be better to be consistent.

We could certainly add a keyword argument to control this behavior, or key modifiers (e.g., select-zoom is ctrl-leftclick-drag rather than just leftclick-drag). Or leave all clickable interaction off by default and document how you add it. Most plotting frameworks seem to have a toolbar, and zoom/pan gets turned on only if you click the appropriate icon in the toolbar. In a multi-axis plot it seems a little weird to not have per-axis controls, but a single figure-global toolbar seems a lot more practical.

jkrumbiegel:
Yes you're right about it being a documented feature. I think some of these differences stem from me developing MakieLayout as my vision of using Makie, so I didn't follow certain patterns that existed in the base library. For example that `lines!(ax, ...)` returns a Lines object and not LAxis or Scene. So now that the two libraries are merged more and more, we need to figure out which solutions are better. Certainly the interaction model has its problems, so we could prototype some solutions in this space.

timholy:
A few growing pains do seem both evident and inevitable. I agree with you about `lines!`. I also think that for 2d plots, the zoom behavior with `LAxis` is closer to what people expect, and it makes sense to consider making this the default.

jkrumbiegel:
Yes mostly the complexity. One thing for example that has changed is that this is how a new LAxis prints now:

```julia
julia> LAxis(scene)
LAxis with 1 plots:
 ┗━ Poly{Tuple{Vector{Vector{Point{2, Float32}}}}}
```

That could be changed to disregard certain plots within `ax.scene` of course, but to me it just feels cleaner to have temporary changes that you can add / remove. Removing things in general I feel is often overlooked but really powerful once you get into more complex GUI scenarios, which are currently quite terrible (removing and readding axes for example leaves so much garbage lying around)

timholy:
So, a short-term strategy would be to move https://github.com/JuliaPlots/AbstractPlotting.jl/blob/5670fe2d9d9b1d9da8f250ed6ab659c916aef5a0/src/makielayout/lobjects/laxis.jl#L724-L729 to a convenience method `select_zoom!(ax)` and document it. I'm happy PR that if you'd like.

piever:
> We could certainly add a keyword argument to control this behavior, or key modifiers (e.g., select-zoom is ctrl-leftclick-drag rather than just leftclick-drag).

As a reference point, Gadfly uses shift-leftclick-drag for select-zoom, see [docs](http://gadflyjl.org/dev/tutorial/#Interactivity-1); just leftclick-drag there is used for panning. So maybe we could use the shift modifier for consistency.

SimonDanisch:
I'd like to change the keyboard defaults!

>Would new users be confused where that plot comes from and what it is?

If printing is the only concern, I'd much rather have a way to not print invisible plots...
In terms of performance, adding & tearing down for an action is so much slower than hiding/unhiding a plot.

Of course, in general we want to make this possible, but for in-built plots like a drag selection, I'd like to use the fastest way possible in the future.

timholy:
Yes, ImageView also uses a modifier (Ctrl-Left-Drag if memory serves). I have no clue what's more common, so either choice seems fine to me.

PR #525, opened 2020-10-18
Update to GridLayoutBase 0.4

timholy:
Closes #519


timholy:
#520 now incorporates the same changes, so probably best to just merge that.

PR #526, opened 2020-10-18
split out recipes into separate files

jkrumbiegel:
I think the previous state was really messy and inconsistent with a couple of separate recipes in files and most others crammed into one.


jkrumbiegel:
I actually found out about a couple of recipes I had never seen before through this haha

PR #527, opened 2020-10-18
Clearer docstring for `rotate_cam!`

ali-ramadhan:
Initially tried calling `rotate_cam!(scene::Scene, 0.1)` when I meant `rotate_cam!(scene::Scene, (0.1, 0, 0))` so just opening a PR to make the docstring a bit clearer.

Wanted to say pitch, yaw, roll but wasn't sure which angled corresponded to which haha.


SimonDanisch:
Thank you!

PR #528, opened 2020-10-18
Update and improve `zoom!` docstring

ali-ramadhan:
Docstring was out of date so I updated it and added a note on `zoom_step`. Couldn't figure out what `shift_lookat::Bool` does so I just included it in the docstring.

### DIFF
--- a/src/camera/camera3d.jl
+++ b/src/camera/camera3d.jl
@@ -215,9 +215,10 @@ function translate_cam!(scene::Scene, cam::Camera3D, _translation::VecTypes)
 end
 
 """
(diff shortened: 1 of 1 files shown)


ffreyer:
It's pretty much what the comment here describe.
```julia
if shift_lookat
    # This results in the point under the cursor remaining stationary
    ...
end
```
I think it would be more precise to say that the passed `point` remains constant here. In the other case the `lookat` is kept constant, so the camera may also rotate. It's more or less the difference between the cad camera and the turntables camera.

PR #529, opened 2020-10-19
add matcap docs

ffreyer:
Documentation for https://github.com/JuliaPlots/GLMakie.jl/pull/125
Also added new default for https://github.com/JuliaPlots/GLMakie.jl/pull/130 so that it's a bit more visible.

I'm also thinking about simplifying the SSAO and lighting examples, should I do this here?

### DIFF
--- a/docs/src/lighting.md
+++ b/docs/src/lighting.md
@@ -33,100 +33,38 @@ A matcap (material capture) is a texture which is applied based on the normals o
 ## Examples
 
 ```@example 1
-using FileIO, MakieGallery, Makie
+xs = -10:0.1:10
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Cool! Whatever works best for you ;)

ffreyer:
Ok, this can be merged then

jkrumbiegel:
Just as an FYI, this PR introduced warnings in the docs build. This is generally something one has to check manually in the gitlab logs when changing docs, because the build does't fail when examples couldn't be rendered. Those just cause warnings. I guess we could change that behavior, but often it seems useful to preview even partially rendered docs.

PR #530, opened 2020-10-19
Fix problems with empty plots / strings / annotations / tick labels

jkrumbiegel:
The most important change is that `broadcast_foreach` does zero iterations if one element is zero length. So far, it always tried one iteration, which would fail with zero-length containers like empty annotations etc.

Then a couple places needed guards against the NaN limits that come from such empty plots usually. In the course of that, LAxis only accepts visible non NaN limits to contribute to autolimits.

`text` also couldn't deal with empty strings because of layouting issues, which it now can. LAxis checks against empty titles and xlabels / ylabels and can now receive empty strings for those attributes.

### DIFF
--- a/src/layouting/layouting.jl
+++ b/src/layouting/layouting.jl
@@ -75,6 +75,8 @@ end
 
 function glyph_positions(str::AbstractString, font_per_char, fontscale_px, halign, valign, lineheight_factor, justification)
 
+    isempty(str) && return Vec2f0[]
+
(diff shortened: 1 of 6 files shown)


jkrumbiegel:
should fix the following:
https://github.com/JuliaPlots/Makie.jl/issues/732
https://github.com/JuliaPlots/Makie.jl/issues/731
https://github.com/jkrumbiegel/GridLayoutBase.jl/issues/5

SimonDanisch (approved):
Awesome! :)

PR #532, opened 2020-10-20
fix Volume #659

jw3126:
close https://github.com/JuliaPlots/Makie.jl/issues/659


test/runtests.jl
```diff
@@ -7,6 +7,14 @@ using GeometryBasics: Pyramid
 using PlotUtils
 using MeshIO, FileIO, AbstractPlotting.MakieLayout
 
+@testset "#659 Volume errors if data is not a cube" begin
+    vol = volume(1:8, 1:8, 1:10, rand(8, 8, 10))
+    lims = AbstractPlotting.data_limits(vol[1])
+    lo, hi = extrema(lims)
+    @test all(lo .<= 1)
+    @test all(hi .>= (8,8,10))
+end
```
jw3126, line 16: Is this there a better way to test this? Also in which file should this test live?

src/layouting/data_limits.jl
```diff
@@ -95,7 +95,9 @@ function data_limits(x::ImageLike{<: Tuple{X, Y, Z}}) where {X, Y, Z}
 end
 
 function data_limits(x::Volume)
-    xyz_boundingbox(transform_func(x), to_value.((x[1], x[2], x[3]))...)
+    _to_interval(r) = ((lo, hi) = extrema(r); lo..hi)
+    axes = (x[1], x[2], x[3])
+    xyz_boundingbox(transform_func(x), _to_interval.(to_value.(axes))...)
```
jw3126, line 100: This is correct for affine transforms. It is incorrect for some nonlinear transforms. I guess all transforms are affine?

SimonDanisch:
Thank you! This is a bit weird, since I'm pretty sure this was working before...I wonder when and why this stopped working!

PR #533, opened 2020-10-22
Zoomlock select

grero:
This is quick and dirty attempt at adding restricted horizontal/vertical zoom when using rubber band selection on LAxis objects.


jkrumbiegel:
this is a good idea and I will add something like it in the course of https://github.com/JuliaPlots/AbstractPlotting.jl/pull/538

jkrumbiegel:
added something like this with `xzoomrect` and `yzoomrect` attributes in the last release :)

grero:
Thanks!

PR #535, opened 2020-10-23
Fix `delete!` bug

ericphanson:
If I try to delete a legend, I get
```julia
MethodError: no method matching hasfield(::Type{AbstractPlotting.MakieLayout.LLegend}, ::Symbol)

You may have intended to import Base.hasfield

delete!(::AbstractPlotting.MakieLayout.LLegend)@helpers.jl:322
top-level scope@Local: 10
```
And if I monkey-patch AbstractPlotting.MakieLayout via
```julia
@eval MakieLayout begin
	function Base.delete!(lobject::Union{LObject, LAxis})
		for (_, d) in lobject.decorations
			remove_element(d)
		end

		if Base.hasfield(typeof(lobject), :scene)
			delete_scene!(lobject.scene)
		end

		GridLayoutBase.remove_from_gridlayout!(GridLayoutBase.gridcontent(lobject))
		return nothing
	end
end
```
it works.

### DIFF
--- a/src/makielayout/helpers.jl
+++ b/src/makielayout/helpers.jl
@@ -319,7 +319,7 @@ function Base.delete!(lobject::Union{LObject, LAxis})
         remove_element(d)
     end
 
-    if hasfield(typeof(lobject), :scene)
+    if Base.hasfield(typeof(lobject), :scene)
         delete_scene!(lobject.scene)
     end
 


ericphanson:
Bump

SimonDanisch:
Sorry, it looked like the wrong fix but I didn't have time to validate it.
But I'm pretty sure, that we shouldn't overshadow `Base.hasfield` in https://github.com/JuliaPlots/AbstractPlotting.jl/blob/5670fe2d9d9b1d9da8f250ed6ab659c916aef5a0/src/makielayout/lobjects/lobject.jl#L9
`hasfield` isn't used anywhere else it seems, so your fix works, but this makes our code base harder to work with, since it arbitrarily "removes" functionality from Base. I think the better fix would be to renamce `hasfield` in `lobjects.jl:9`

ericphanson:
Ah, that makes sense, I'll open a new PR with that fix.

PR #536, opened 2020-10-24
Added option of toggling rubber-band zooming

vilim:
The currently-default, always-on rubber band LAxis zoom (introduced in a recent commit) breaks many interactive applications. I added an attribute to be able to disable it. It involves a slightly inelegant addition in `select_rectangle`, but I think it's a working stopgap solution until a more comprehensive framework for interactive tools is implemented (akin to matplotlib or plotly's tool bar)

### DIFF
--- a/src/interaction/interactive_api.jl
+++ b/src/interaction/interactive_api.jl
@@ -146,7 +146,7 @@ rectangle has area > 0.
 
 The `kwargs...` are propagated into `lines!` which plots the selected rectangle.
 """
-function select_rectangle(scene; strokewidth = 3.0, kwargs...)
+function select_rectangle(scene, enabled=Node(true); strokewidth = 3.0, kwargs...)
     key = Mouse.left
     waspressed = Node(false)
     rect = Node(FRect(0, 0, 1, 1)) # plotted rectangle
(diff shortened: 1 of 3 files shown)


grero:
Interesting, I just found the need for this myself and was about to submit a PR. : )

jkrumbiegel:
yeah I'm not surprised! This was merged a bit too hastily. Actually, with the slight change to Observables recently which allows to disconnect them more easily, I think I want to tackle this problem more cleanly and offer an easy and clean way to connect and disconnect such mouse interactions to and from an axis. I basically want these to be able to change dynamically, so that you can switch between different modes of interaction, or disable them all. I will therefore not merge this particular PR, because that would lead in the direction of a lot of different toggles :) but instead it'll probably be something like `add_interaction_xyz!` and `remove_interaction_xyz!`, with some sort of interface to the mouse observables, so there's not so much duplication in the backend.

jkrumbiegel:
here's my current proposed solution https://github.com/JuliaPlots/AbstractPlotting.jl/pull/538

jkrumbiegel:
the latest release has this now, `register_interaction!`, `deregister_interaction!`, `activate_interaction!` and `deactivate_interaction!`

PR #538, opened 2020-10-27
change mouse events and LAxis interaction backend logic

jkrumbiegel:
The current state is that adding and removing interaction to an axis is not straightforward. This aims to simplify the pattern one needs to implement to safely add and remove or activate / deactivate interaction modes.

First I changed the mouse events to an enum so there's a finite set of them and there's less compilation needed.
Then, the idea is that the axis has a list of `AbstractInteraction`s stored with a bool for active / inactive. Every mouse or scroll event that comes in (so far, keyboard events should be added I guess but are not needed for the current set of interactions) calls `process_interaction(interaction, event, parentobject)`.

This can easily be overloaded for custom `AbstractInteractions`.

New interactions are added with `register_interaction!(parent, name, interaction)` and removed with `deregister_interaction!(parent, name)`. During registration / deregistration one can overload `registration_setup!(parent, interaction)` and `deregistration_cleanup!(parent, interaction)` just to make sure that mutated state can be cleaned up if necessary.

One can also call `activate_interaction!(parent, name)` and `deactivate_interaction!(parent, name)` to leave interactions in place but toggle them on or off. This should be useful to temporarily deactivate rubberband selection for example, if one needs the left mouse drag. One would currently just call `deactivate_interaction!(ax, :rectanglezoom)`.

Here's an example how rectangle zoom is implemented with this scheme, where the rectangle poly is added and removed as needed, keeping the scene structure clean:

```julia
mutable struct RectangleZoom <: AbstractInteraction
    from::Union{Nothing, Point2f0}
    to::Union{Nothing, Point2f0}
    rectnode::Observable{FRect2D}
    poly::Union{Poly, Nothing}
end



function process_interaction(r::RectangleZoom, event::MouseEvent, ax::LAxis)

    if event.type === MouseEventTypes.leftdragstart
        r.from = event.prev_data
        r.to = event.data
        r.rectnode[] = FRect2D(r.to, r.from .- r.to)
        r.poly = poly!(ax.scene, r.rectnode,
            color = (:blue, 0.1), strokewidth = 1, strokecolor = (:blue, 0.5))[end]

    elseif event.type === MouseEventTypes.leftdrag
        r.to = event.data
        r.rectnode[] = FRect2D(r.to, r.from .- r.to)

    elseif event.type === MouseEventTypes.leftdragstop
        newlims = positivize(r.rectnode[])
        if !(0 in widths(newlims))
            ax.targetlimits[] = newlims
        end

        if !isnothing(r.poly)
            delete!(ax.scene, r.poly)
            r.poly = nothing
        end
    end

    return nothing
end



register_interaction!(ax,
        :rectanglezoom,
        RectangleZoom(nothing, nothing, Node(FRect2D(0, 0, 1, 1)), nothing))
```


grero:
This looks really nice! I can't wait to try it out for my interactive plots.

jkrumbiegel:
I've actually simplified a bit more and removed the `AbstractInteraction` thing. Just any object that works with the necessary methods is fine, additionally one can pass Functions which are called directly with event and parent. That should offer the least amount of friction

ffreyer:
It would be nice if you could write `do` blocks as well then. Maybe adding an alias `on_event(ax, :myinteraction) do ... end` would work?

PR #539, opened 2020-10-27
added a visible attribute to BarPlot

kragol:
This just adds a `visible` attribute to `BarPlot`. It looks like some other compound plot types in basic_recipes (e.g. `Contour`?) could use that too but I am not familiar enough with those atm.

Related issue: JuliaPlots/Makie.jl#744


SimonDanisch:
Thank you!
Yeah, visible should be present for all plots!

PR #542, opened 2020-10-29
not necessary to split intersections anymore because of earcut

jkrumbiegel:


jkrumbiegel:
misunderstood what earcut could do, seems like it doesn't handle self intersections after all

SimonDanisch:
https://github.com/mapbox/earcut/issues/95#issuecomment-368442969

PR #543, opened 2020-10-29
Don't shadow `hasfield`

ericphanson:
Alternate fix to https://github.com/JuliaPlots/AbstractPlotting.jl/pull/535 as suggested by @SimonDanisch


ericphanson:
Could you take a look @jkrumbiegel ? Since the bug is in the MakieLayout code

jkrumbiegel:
looks good, thanks! I remember there being something weird going on when I originally had it as Base.hasfield, and then for some reason I changed it to this. Don't quite remember though and it seems to work fine now

ericphanson:
thanks!

PR #544, opened 2020-10-30
Make texture atlas serialization julia-version specific

ianshmean:
Seems like a good idea? Especially for people switching between julia versions


SimonDanisch:
Makes sense, thank you!

PR #549, opened 2020-11-06
fix Makie #749 + docstrings

ffreyer:
https://github.com/JuliaPlots/Makie.jl/issues/749

Currently `ispressed(scene, button)` always returns false when multiple buttons are pressed simultaneously. This pr changes that behavior, so that `ispressed(scene, button)` returns true if `button` and `other_button` are both pressed.

### DIFF
--- a/src/interaction/events.jl
+++ b/src/interaction/events.jl
@@ -91,6 +91,10 @@ function ispressed(buttons::Set{T}, button::T) where T <: Union{Keyboard.Button,
     return button in buttons
 end
 
+function ispressed(buttons::Set{T}, button::Set{T}) where T <: Union{Keyboard.Button, Mouse.Button}
+    return issubset(button, buttons)
+end
+
 """
     ispressed(scene, button)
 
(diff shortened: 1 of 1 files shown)


SimonDanisch:
With this PR, how would you check that exactly only the buttons you query are pressed?

ffreyer:
Manually I guess? But why would you want to check that?

SimonDanisch:
I feel like the default is that I want only one button to be pressed, since pressing multiple buttons usually means something different... But maybe I haven't used it enough yet myself...

ffreyer:
The issue is an example of where you'd want multiple buttons to trigger multiple things, rather than nothing. E.g. up + left should trigger diagonal translation.

I've used QWEASD to do 3D rotations before (i.e. camera rotations, pan, tilt and roll). I did want multiple to trigger at once there too.

If you want to do things like `Ctrl + c` you currently can't do `ispressed(scene, Key.Ctrl) && ispressed(scene, Key.c)` because that'd require pressing two buttons. With this pr you could. Though that would also trigger on `Ctrl + c + a` etc.

Also just by name, `ispressed` should return true whenever the given button is pressed. I don't see why `ispressed` should return false if the given button is pressed alongside another (again, given the name of the function).

SimonDanisch:
Alright, seems convincing ;) Thanks!

PR #550, opened 2020-11-07
Faster recording

ianshmean:
Speeds up recording about 2x in my tests (but the underlying colorbuffer func is 3.5x faster)

Test animation
```julia
function rec()
    points = Node(Point2f0[(0, 0)])
    scene = Scene(resolution = (3000,2000))
    scatter!(scene, points, limits = FRect(0, 0, 30, 30))
    frames = 1:30
    GLMakie.set_window_config!(pause_rendering=true)
    record(scene, "append_animation.mp4", frames; framerate = 30) do frame
        new_point = Point2f0(frame, frame)
        points[] = push!(points[], new_point)
    end
end
```

This PR + https://github.com/JuliaPlots/GLMakie.jl/pull/138

```
julia> @time rec()
  4.848452 seconds (177.27 k allocations: 27.778 MiB)

julia> @btime Makie.AbstractPlotting.colorbuffer(scene, ffmpeg_format = true);
  44.931 ms (568 allocations: 9.58 KiB)
```

Master on AbstractPlotting and GLMakie:
```
julia> @time rec()
  9.264750 seconds (177.36 k allocations: 542.763 MiB, 0.58% gc time)

julia> @btime Makie.AbstractPlotting.colorbuffer(scene);
  156.547 ms (568 allocations: 9.58 KiB)
```


src/display.jl
```diff
@@ -430,16 +432,15 @@ Adds a video frame to the VideoStream `io`.
...
-    _xdim, _ydim = size(frame)
-    xdim = _xdim % 2 == 0 ? _xdim : _xdim + 1
-    ydim = _ydim % 2 == 0 ? _ydim : _ydim + 1
-    frame_out = fill(RGB{N0f8}(1, 1, 1), ydim, xdim)
-    for x in 1:_xdim, y in 1:_ydim
-        c = frame[(_xdim + 1) - x, y]
-        frame_out[y, x] = RGB{N0f8}(c)
+    frame = colorbuffer(io.screen, ffmpeg_format = true)
+    _ydim, _xdim = size(frame)
+    if isodd(_xdim) || isodd(_ydim)
```
ianshmean, line 437: Is it actually possible for this to be true? I tried setting the scene window size to an odd number, but the sizes here always returned even.
If something's coercing the dims to even, then this and paddedview dep can be removed

src/display.jl
```diff
@@ -396,7 +396,7 @@ end
 
 
 # This has to be overloaded by the backend for its screen type.
-function colorbuffer(x)
+function colorbuffer(x; kwargs...)
     error("colorbuffer not implemented for screen $(typeof(x))")
```
SimonDanisch, line 400: can we make the supported kw args explicit with docs? :)

ianshmean: It's listed just below on 411. But I'll add it explicitly everywhere 👍🏻

src/display.jl
SimonDanisch: Good question :D

ianshmean: I wondered if it might be a retina display thing? I only tried it on MacOS

SimonDanisch: Oh yeah, this is very likely platform dependent :D

ianshmean: Cool. It would be interesting to see how much slower recording an odd-sized window would be

SimonDanisch: Let me test today on win & linux... Would be nice to not have additional complexity for something that can't happen :D

ianshmean:
Brought #552 into here and finished off the overloading backwards compat handling

@SimonDanisch does this look right?

ianshmean:
@SimonDanisch I just tested backwards compat and some tweaks were needed. This now works with this PR and GLMakie `v0.1.14`

SimonDanisch:
Thanks for making it backwards compatible :) 
But I do get this method error:
```
no method reverse!(::Array{ColorTypes.RGB{FixedPointNumbers.Normed{UInt8,8}},2}; dims=1)
```

ianshmean:
Updated for support for <1.6 julia, given `reverse!(A, dims)` was added in 1.6

## Julia 1.6
This PR + https://github.com/JuliaPlots/GLMakie.jl/pull/138
```
julia> @time rec()
  4.788138 seconds (177.45 k allocations: 27.783 MiB, 0.07% compilation time)
```
This PR + GLMakie 0.1.14
```
julia> @time rec()
┌ Warning: Inefficient re-conversion back to GLNative buffer format. Update GLMakie to support direct buffer access
└ @ AbstractPlotting ~/.julia/dev/AbstractPlotting/src/display.jl:407
 11.199670 seconds (484.92 k allocations: 1.050 GiB, 0.88% gc time, 1.22% compilation time)
```

## Julia 1.5
This PR + https://github.com/JuliaPlots/GLMakie.jl/pull/138
```
julia> @time rec()
  4.945769 seconds (137.49 k allocations: 27.390 MiB)
```
This PR + GLMakie 0.1.14
```
julia> @time rec()
┌ Warning: Inefficient re-conversion back to GLNative buffer format. Update GLMakie to support direct buffer access
└ @ AbstractPlotting ~/.julia/dev/AbstractPlotting/src/display.jl:407
 11.650214 seconds (137.80 k allocations: 1.033 GiB, 1.98% gc time)
```

PR #559, opened 2020-11-21
[WIP] Update interaction functions select_thing to new registration pipeline

Datseris:
I'm taking baby steps here because I'm a new comer. The goal of this PR is to slowly update the functions `select_line, select_rectangle, select_point`, that I've contributed some time ago, to the new registration interaction API that Julius wrote.

To my knowledge these functions are used in InteractiveChaos.jl, but I don't think they're used anywhere else.

At the moment I'm starting with `select_rectangle` which was very easy to make it work. Now I need to make it properly implement the registration API.

### DIFF
--- a/src/makielayout/interactions.jl
+++ b/src/makielayout/interactions.jl
@@ -109,11 +109,11 @@ function _chosen_limits(rz, ax)
     lims = ax.limits[]
     # restrict to y change
     if rz.restrict_x || !ax.xrectzoom[]
-        r = FRect2D(lims.origin[1], r.origin[2], widths(lims)[1], widths(r)[2]) 
+        r = FRect2D(lims.origin[1], r.origin[2], widths(lims)[1], widths(r)[2])
     end
     # restrict to x change
     if rz.restrict_y || !ax.yrectzoom[]
-        r = FRect2D(r.origin[1], lims.origin[2], widths(r)[1], widths(lims)[2]) 
+        r = FRect2D(r.origin[1], lims.origin[2], widths(r)[1], widths(lims)[2])
     end
     return r
 end
@@ -298,6 +298,14 @@ function process_interaction(dp::DragPan, event::MouseEvent, ax)
     timed_ticklabelspace_reset(ax, dp.reset_timer, dp.prev_xticklabelspace, dp.prev_yticklabelspace, dp.reset_delay)
 
     tlimits[] = FRect(Vec2f0(xori, yori), widths(tlimits[]))
-           
+
     return nothing
+end
+
+############################################################################
+#             Interactions that return values when activated               #
+############################################################################
+function AbstractPlotting.select_rectangle(ax::MakieLayout.LAxis; kwargs...)
+    deactivate_interaction!(ax, :rectanglezoom)
+    select_rectangle(ax.scene; kwargs...)
 end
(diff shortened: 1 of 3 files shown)


Datseris:
@jkrumbiegel I think I know how to proceed here. For the `select_rectangle` function I need to create a type `SelectRectangle`, and then essentially copy your code:

```julia
function process_interaction(r::RectangleZoom, event::MouseEvent, ax::LAxis)
```
to dispatch on my type. Then the function `select_rectangle` only needs to initialize the struct, register it on the axis, and make sure to return the rectangle observable of interest, which is in fact `r.rectnode`.

One thing I'm not sure is why this function's plotting code is so much different than then original source of `select_rectangle`. Is it better to plot, delete, and replot the selected rectangle, instead of the approach I used where I did `plotted_rect[:visible] = false` instead?

Datseris:
At the moment everything works fine. I haven't yet updated the interaction to the new pipeline with registrations and all, but they work never the less. I need to be using InteractiveChaos.jl in an upcoming course I give next Wednesday, so can you please merge and tag a new version with this PR in, so the students can use it? Then I can work on updating the interface when I have more time.

src/makielayout/interactions.jl
```diff
@@ -298,6 +298,28 @@ function process_interaction(dp::DragPan, event::MouseEvent, ax)
...
-           
+
     return nothing
 end
+
+############################################################################
+#             Interactions that return values when activated               #
+############################################################################
+function AbstractPlotting.select_rectangle(ax::MakieLayout.LAxis; kwargs...)
+    deactivate_interaction!(ax, :rectanglezoom)
```
jkrumbiegel, line 309: I don't think interactions should deactivate others by default. I'm not sure what the best solution is, but there could always be other interactions that are enabled, which could conflict. So the person who enables one interaction needs some knowledge about what they disable instead, I think.

jkrumbiegel: Rectanglezoom could also theoretically be set to a different non-conflicting mouse button, etc.

Datseris: > I don't think interactions should deactivate others by default.

It doesn't have to be by default, it can be a keyword option to `select_rectangle`. Of course I agree that the developer that uses `select_rectangle` (or whatever interaction) should be aware of what interactions are present.

> Rectanglezoom could also theoretically be set to a different non-conflicting mouse button, etc.

I disagree, because the current amount of keys/buttons taken by the default interaction on `LAxis` is simply too much, leaving no sensible options. Many people have mouses with only two buttons (anything else besides a mouse button doesn't feel intuitive for `interactive_rectangle`). At the moment by default both left and right clicks are taken. Besides, there is already the scroll wheel available for zooming, this (in my eyes) only enhances the argument that left click should be allowed to be rebinded, given that zooming possibility isn't actually lost.

Datseris: I now realize that a sensible option would be the case of having to press Ctrl first (but I don't know how to do the proper checks for this)

Datseris: on the other hand one can also consider letting Left mouse button be free, the option to be used by front-end interactive applications, and make the zoom rectangle require control

jkrumbiegel:
I'm going to close this as we talked about needing a different implementation on slack and we're trying to clean up PRs currently. Feel free to reopen anytime you think you want to revisit the topic

PR #560, opened 2020-11-24
fix Axis3D scaling issues

ffreyer:
Example from Slack:

```julia
using AbstractPlotting, GLMakie
x = rand(100).*260.0;
y = rand(100).*10000.0;
z = rand(100).*7400.0;
maxvals = map(a->maximum(a), [x,y,z]);
maxmax = maximum(maxvals);
scale_factors = map(a->maxmax/a, maxvals);
s2 = AbstractPlotting.Scene();
AbstractPlotting.scatter!(s2,x,y,z,markersize=500);
axes=s2[Axis];
axes[:names][:axisnames]=("East-west","North-south","Up down");
axes[:names][:textsize]=(200,200,200);
axes[:ticks][:textsize]=(120,120,120);
AbstractPlotting.scale!(s2,scale_factors[1],scale_factors[2],scale_factors[3]);
display(s2)
```

produces

image

This pr fixes two things - the spacing of axis labels (via the `w / scale[j]`) and the gap between ticks (numbers) and axis lines in scaled plots. 

With this pr:

Screenshot from 2020-11-24 22-38-53

### DIFF
--- a/src/basic_recipes/axis.jl
+++ b/src/basic_recipes/axis.jl
@@ -601,7 +601,7 @@ to3tuple(x::Tuple{Any, Any}) = (x[1], x[2], x[2])
 to3tuple(x::Tuple{Any, Any, Any}) = x
 to3tuple(x) = ntuple(i-> x, Val(3))
 
-function draw_axis3d(textbuffer, linebuffer, limits, ranges_labels, args...)
+function draw_axis3d(textbuffer, linebuffer, scale, limits, ranges_labels, args...)
     # make sure we extend all args to 3D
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Thanks! I'm surprised, that the text looks so bad in the second picture... I guess this can't be related to this PR though!

ffreyer:
Yea, I don't see why this would change text rendering. Maybe it's just a resolution thing?
From my side the pr is done btw

SimonDanisch:
Thanks :)

PR #563, opened 2020-11-30
Change _primary_resolution to use system calls on Apple computers

ndinsmore:
This fixes primary_resolution() so that it returns that actual current display setting.   It also changes the call to use system calls which should be faster.

### DIFF
--- a/src/theming.jl
+++ b/src/theming.jl
@@ -8,10 +8,12 @@ if Sys.iswindows()
         end
     end
 elseif Sys.isapple()
+    const _CoreGraphics = "/System/Library/Frameworks/CoreGraphics.framework/CoreGraphics"
     function _primary_resolution()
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Awesome, thanks a lot! :) 
Do you happen to know, how portable this is between OSX versions, considering the `_CoreGraphics` path is hardcoded?

ndinsmore:
Simon CoreGraphics has been a core framework since 10.8. and aple list support as 10.8+.  I changed the library string to allow the system to find the path.   Should be portable now.

SimonDanisch:
Cool, whatever works! ;) I just have no way to test this, and the CI neither!

ndinsmore:
I think that this is ready to go.

SimonDanisch:
Thanks!

PR #564, opened 2020-12-02
Jk/errorbars improvement

jkrumbiegel:
use a more sparse three-value representation (the value between high and low is useless), use :horizontal and :vertical for clarity and allow dynamic resizing through using either a vector of triplets, or because the other path goes through conversion now and can use the `x.val = ...; low.val = ...; high[] = newvector` approach


jkrumbiegel:
also the data limits correctly ignore the whiskers now (which are not meaningful in data space)

src/basic_recipes/errorbars.jl
```diff
@@ -17,60 +16,44 @@ $(ATTRIBUTES)
...
+        if dir == :vertical
             true
-        elseif dir == :x
+        elseif dir == :horizontal
             false
         else
-            error("Invalid direction $dir. Options are :x and :y.")
+            error("Invalid direction $dir. Options are :horizontal and :vertical.")
         end
     end
```
ffreyer, line 48: I find `:x` and `:y` easier to remember, maybe we could allow both?

src/basic_recipes/errorbars.jl
```diff
@@ -1,13 +1,12 @@
 """
-    errorbars(xs, ys, low, high; kwargs...)
-    errorbars(xs, ys, lowhigh; kwargs...)
-    errorbars(points, low, high; kwargs...)
-    errorbars(points, lowhigh; kwargs...)
+    errorbars(pos, low, high; kwargs...)
```
ffreyer, line 2: For me the most common case (well actually the only) is that `high = ys .+ deltas` and `low = ys .- deltas`. It would be nice if I could just do `errorbars(xs, ys, deltas)`.

Maybe we could have:
```julia
errorbars(xs::Vector{Float}, ys::Vector{Float}, deltas::Vector{Float})
errorbars(xs::Vector{Float}, low_high::Vector{Point2f0})
errorbars(xys::Vector{Point2f0}, deltas::Vector{Float})
```

Also `pos` isn't very clear, it makes me think of `(x, y)` positions. Just `xs` would be more clear imo.

ffreyer (changes requested):

src/basic_recipes/errorbars.jl
ffreyer: I guess `xs` isn't great either since those are either x or y values. Maybe writing out the type would clear this up?

src/basic_recipes/errorbars.jl
jkrumbiegel: I actually changed it back just now because I came to the same conclusion after changing xs to pos

src/basic_recipes/errorbars.jl
jkrumbiegel: I guess the `errorbars(xys::Vector{Point2f0}, deltas::Vector{Float})` signature could work, the first one conflicts with the normal definition, which I'd like to keep as the default

jkrumbiegel: the second one could also work

jkrumbiegel: It's common to have only one delta for standard deviations etc, but common to have two separate values for any kind of bootstrapping

ffreyer: I was thinking of the second one as an alternative to what you have now. I never needed different deltas for low and high, so I'd prefer having `xs, ys, deltas`. Otherwise I'd always have to do the extra work of `lows = ys .- deltas; highs = ys .+ deltas` before plotting.

src/basic_recipes/errorbars.jl
ffreyer: I don't think it would hurt to just have `dir in (:y, :vertical)` for the people that find that more intuitive

ffreyer:
I also wonder whether `pos_low_high` should be a `Vector{Point3f0}`. That seems more in line with the usual interface we have, but might be confusing since it doesn't specify an `(x, y, z)` point.

src/basic_recipes/errorbars.jl
jkrumbiegel: yeah that just depends I guess, I mostly need low and high :)

maybe we can do it this way, with a tuple of vectors option for high low (because it's also a bit annoying having to make points from them):

```julia
# x and y should not flip with direction?
errorbars(x::Vector{Float}, y::Vector{Float}, delta::Vector{Float})

# here val would be either x or y given the direction
errorbars(val::Vector{Float}, high_low::Tuple{Vector{Float}, Vector{Float}})
errorbars(val::Vector{Float}, high_low::Vector{Point2})

# xy should not flip with direction?
errorbars(xy::Vector{Point2}, delta::Vector{Float})
errorbars(xy::Vector{Point2}, high_low::Tuple{Vector{Float}, Vector{Float}})
errorbars(xy::Vector{Point2}, high_low::Vector{Point2})
```

jkrumbiegel: although I don't like so much that there is `delta` in one, and `high_low` in the other

jkrumbiegel:
> I also wonder whether pos_low_high should be a Vector{Point3f0}

It could be that or a tuple of three values, both would work

src/basic_recipes/errorbars.jl
jkrumbiegel: maybe we need two recipes then, one for delta and one for high/low? maybe `errorbars` for the delta version and `errorranges` for the high-low version

jkrumbiegel: or `errorbars` and `rangebars`

ffreyer: I would agree with keeping the xy order regardless of `dir`.

> maybe we need two recipes then, one for delta and one for high/low?
That would work as well.

ffreyer:
Oh, ok, nevermind. I thought the conversion was only `Tuple -> Point`

ffreyer (approved):
Looks good to me :+1:

PR #567, opened 2020-12-05
convert colorbuffer frames to rgb and fix out path

jkrumbiegel:
this was for the messed up cairo videos that use an rgba buffer


jkrumbiegel:
@SimonDanisch this still causes warnings like:

```julia
┌ Warning: Inefficient re-conversion back to GLNative buffer format. Update GLMakie to support direct buffer access
└ @ AbstractPlotting /builds/JuliaGPU/AbstractPlotting-jl/src/display.jl:410
```

what should we do about those?

PR #571, opened 2020-12-10
Update basic-tutorial.md

bvilmann:
I was going through the tutorial and found some issues:

1) `AbstractPlotting.inline!(true)` requires `AbstractPlotting` as is.

2) `markersize` is not referring to pixels by default even though the [docs](http://makie.juliaplots.org/stable/generated/plot-attributes.html) inclines. This attribute has to be defined like `markersize = 15px`. [This issue](https://github.com/JuliaPlots/Makie.jl/issues/772) also shows that a markersize of mere 0.7 plots oversized points relative to pixels.

All the best!
B

PS: I am new to github so please let me know if I made a wrong procedure!

### DIFF
--- a/docs/src/basic-tutorial.md
+++ b/docs/src/basic-tutorial.md
@@ -10,9 +10,10 @@ nothing # hide
 ```
 
 For this tutorial, we also call `AbstractPlotting.inline!(true)` so plots appear inline after each example.
-Otherwise, an interactive window will open when you return a `Scene`.
+Otherwise, an interactive window will open when you return a `Scene`. To enable this, we need to import `AbstractPlotting`.
 
 ```@example 1
+using AbstractPlotting
 AbstractPlotting.inline!(true)
 nothing # hide
 ```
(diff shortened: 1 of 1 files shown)


SimonDanisch:
>This attribute has to be defined like markersize = 15px

Oh interesting, that must be some weird regression - markersize should definitely default to pixels... I'll investigate when I have some time

bvilmann:
Yeah! It is interesting what unit it is scaling to.

However, I actually kind of liked the approach that you can scale your plots to whatever unit like in `LaTeX` :)

Cool.

SimonDanisch:
The plot thickens, I think you're on an old version of AbstractPlotting, where pixel units haven't been the default ;) 
You can still set the units, but currently we only have pixel / scenespace:
```julia
scatter(1:4, markersize=0.1, markerspace=SceneSpace) # units are in the same as x,y - some people also call this axis space
```
More units aren't supported right now, but are planned to be supported ;)
Thanks for fixing instead of opening an issue! :) I'll still close for now, since there doesn't seem to be any problems with the docs.
I guess `using AbstractPlotting` could be added, maybe in another PR?

PR #578, opened 2020-12-26
Add a whole bunch of low-quality precompiles

timholy:
The really useful part of this is the specialization/inferrability work, but until people have time for more of it (xref https://github.com/JuliaPlots/Makie.jl/issues/792) then the heap of automatically-generated precompiles is probably about the best we can do. Together with PRs in other repositories, this shaves something like 6s off TTFP for me.


timholy:
Requires https://github.com/JuliaGeometry/GeometryBasics.jl/pull/115

PR #579, opened 2020-12-27
Docs for setting "read-only" attributes & `set_range!` function

yha:
Closes JuliaPlots/Makie.jl#794 by updating docs of relevant attributes. 
Also adds `set_range!` to fix a similar issue with setting a slider's range.


jkrumbiegel:
In the last version I fixed the underlying issue with the `range` attribute and you can now change it directly, which will update the slider to the closest value in the new range. Thanks for making me aware of the issue, I'll close this PR now.

PR #581, opened 2020-12-28
Make button color updates take effect immediately

yha:
This makes update such as `button.buttoncolor = ...` show immediately rather than on the next mouse event on the button.
Demo script for testing:
```
scene, layout = layoutscene()
default_sl = layout[1,1] = LSlider(scene; range=0:0.1:1)

b = layout[2,1] = LButton(scene)
on(default_sl.value) do v
    b.buttoncolor = Gray(v)
    b.labelcolor = Gray(1-v)
end
on(scene.events.keyboardbuttons) do button
    if ispressed(button, Keyboard.space)
        for sym in (:buttoncolor_hover, :buttoncolor_active, :labelcolor_hover, :labelcolor_active)
            b.:($sym) = rand(RGBf0)
        end
    end
end
scene
```


jkrumbiegel:
thanks, looks good!

PR #586, opened 2021-01-03
Events refactor

ffreyer:
### TODO

- [x] add new event types
- [x] add `Interactions` (which holds and organizes objects to act on events)
- [x] add input state (what keys are pressed, etc)
- [x] add new event system
- [ ] QoL stuff (show, getters, ...)
- [x] update camera controllers (prototype, need to decide what to do exactly)
- [x] update standard Makie interactions
- [ ] update all event based Nodes
- [ ] remove or deprecate MakieLayout interactions
- [x] add new `mousestatemachine` and deprecate or remove the old one (prototype)
- [ ] update layoutables
- [ ] add tests
- [ ] add documentation
- [ ] add deprecations
- [ ] cleanup

---

This is an attempt to rework the event system to allow for blocking events. Having blocking events should fix https://github.com/JuliaPlots/Makie.jl/issues/777 for example.

The basic idea here is the following:
* The backend generates an event, e.g. `KeyEvent(key, action)` and calls `process(top_level_scene, event)`
* The event is passed down to all child scenes and plots, in order of priority (first) and render order (second)
* The event is then passed on to interactions (functions or objects) where they maybe consumed by returning `true`

Interactions are similar to https://github.com/JuliaPlots/AbstractPlotting.jl/pull/538. They can be `de/register!`ed to a scene or plot and are processed by `process!(interaction, event, parent)` or by calling `function_interaction(event, parent)`. My version also has a priority attached to it, which controls the order at which interactions process events. 

At the current state of this only the text input of `LTextbox` and window resizing uses the new system. I wanted to get some feedback before I start translating things from the old system. (This should be able to replace `scene.events` completely if `ispressed` querries the backend instead. It should also be able to replace MakieLayout interactions.) Is this something I should continue working on?

A working example with blocking text input: (requires GLMakie branch: https://github.com/JuliaPlots/GLMakie.jl/pull/154)
```julia
scene, layout = layoutscene()
a = layout[1, 1] = LAxis(scene)
b = layout[2, 1] = LTextbox(scene, tellwidth=false)
points = Node(rand(Point2f0, 10))
p = scatter!(a, points)
# default priority, any event type
register!(p, :test) do event, p
    points[] = rand(Point2f0, 10)
    false # does not consume events
end
scene
```


ffreyer:
Something this would enable are [interactive Gizmos](https://github.com/CedricGuillemet/ImGuizmo) to manipulate 3D objects/plots. On the master branch Gizmos like these would trigger camera updates as well. Here one can create a higher priority interaction that consumes mouse events before they reach the camera, allowing this to work as expected.

Here's an example translation gizmo which should work with any 3D plot object with a model matrix. You can mouse drag the different pieces around to get motion in x/y/z direction, xy/yz/xz planes or in the screen plane.

https://user-images.githubusercontent.com/10947937/104129413-70dc6f00-536c-11eb-9f6d-dd3d6657ab61.mp4

https://gist.github.com/ffreyer/5bb087ba94fc95f9c9607db7deb63db7

ffreyer:
I've done some performance testing, stared into the distance for a while and decided I want to redo this with a different approach. I'll close this pr and open a new one.

For performance testing I checked a grid of 11x12 Toggles. Comparing actions with immediate returns, the new version takes about 800µs and observables only ~40µs. This is still ok-ish, but it'll slow down further with subscenes, more plot objects and more unique priorities. I'm worried this may become too slow in advanced applications.

My new approach is to make a new observable type which includes priorities and early exits. With that we can keep the `Events` struct and can keep using `on` though the observer functions must return a Bool and `map` is not allowed. With that the immediate return case is just as fast as with normal observables and it should scale better with the number of scenes, plots and priorities.

PR #590, opened 2021-01-06
Precompilation docs

pbouffard:
It looks like the bit about precompilation using PackageCompiler didn't get transferred from Makie.jl to AbstractPlotting.jl. This PR just copies that onto a separate page in the docs and links there from the home page.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -367,6 +367,7 @@ makedocs(
             "backends.md",
             "troubleshooting.md",
             "API Reference AbstractPlotting" => "abstractplotting_api.md",
+            "precompilation.md",
         ],
         "MakieLayout" => [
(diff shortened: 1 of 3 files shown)


pbouffard:
Hold up -- I didn't at first notice that the precompilation script itself hadn't been ported over and now that I'm trying to use it I'm not sure it works without modification. Closing the PR while sorting that out.

SimonDanisch:
Yeah I need to update those at some point :D

pbouffard:
So what's the conventional wisdom at this point? Is there at least a minimal precompilation script that will speed up initial Makie import? Most of the time I can avoid delays using Revise but as soon as I modify a struct, that requires a Julia restart.

PR #593, opened 2021-01-07
improve linestyles, add "-.-." and (:dash, :loose) variants

greimel:
This improves CairoMakie. Adjusting the other two backends is still todo.

Before:

image

After:

image

```julia
using CairoMakie

fig, ax, plt = lines(1:2, 2:3, linestyle=:dash)
lines!(ax, 1:2, 3:4, linestyle=:dashdot)
lines!(ax, 1:2, 4:5, linestyle=:dot)
lines!(ax, 1:2, 5:6, linestyle=:dashdotdot)

lines!(ax, 2:3, 2:3, linestyle=:dash, linewidth=5)
lines!(ax, 2:3, 3:4, linestyle=:dashdot, linewidth=5)
lines!(ax, 2:3, 4:5, linestyle=:dot, linewidth=5)
lines!(ax, 2:3, 5:6, linestyle=:dashdotdot, linewidth=5)

fig
```

fixes https://github.com/JuliaPlots/Makie.jl/issues/807

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -741,10 +741,8 @@ end
 
 "Checks if the linestyle format provided as a string contains only dashes and dots"
 function check_pattern(ls_str)
-    for i in 1:length(ls_str)
-        ls_str[i] == '-' || ls_str[i] == '.' ||
+    isnothing(match(r"^[\.\-]+$", ls_str)) &&
         throw(ArgumentError("If you provide a string as linestyle, it must only consist of dashes (-) and dots (.)"))
-    end
 end
 
 function convert_gaps(gaps)
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -680,19 +680,8 @@ convert_attribute(A::AbstractVector, ::key"linestyle") = A
 function convert_attribute(ls::Symbol, ::key"linestyle")
     return if ls == :solid
         nothing
-    elseif ls == :dash
-        [0.0, 1.0, 2.0, 3.0, 4.0]
-    elseif ls == :dot
-        tick, gap = 1/2, 1/4
-        [0.0, tick, tick+gap, 2tick+gap, 2tick+2gap]
-    elseif ls == :dashdot
-        dtick, dgap = 1.0, 1.0
-        ptick, pgap = 1/2, 1/4
-        [0.0, dtick, dtick+dgap, dtick+dgap+ptick, dtick+dgap+ptick+pgap]
-    elseif ls == :dashdotdot
-        dtick, dgap = 1.0, 1.0
-        ptick, pgap = 1/2, 1/4
-        [0.0, dtick, dtick+dgap, dtick+dgap+ptick, dtick+dgap+ptick+pgap, dtick+dgap+ptick+pgap+ptick,  dtick+dgap+ptick+pgap+ptick+pgap]
+    elseif ls in [:dash, :dot, :dashdot, :dashdotdot]
+        pattern(ls)
     else
         error(
             """
@@ -702,6 +691,32 @@ function convert_attribute(ls::Symbol, ::key"linestyle")
             This sequence of numbers must be cumulative; 1 unit corresponds to 1 line width.
             """
         )
+    end
+end
+
+function pattern(linestyle, modifier=:normal; kwargs...)
+    float.([0; cumsum(_pattern(linestyle, modifier; kwargs...))])
+end
+
+"The linestyle patterns are inspired by the LaTeX package tikZ as seen here https://tex.stackexchange.com/questions/45275/tikz-get-values-for-predefined-dash-patterns."
+function _pattern(linestyle, modifier=:normal;
+                  dot = 1,  dot_gap  = (normal = 2, dense = 1, loose = 4),
+                  dash = 3, dash_gap = (normal = 3, dense = 2, loose = 6)
+                  )
+
+    if linestyle == :dash
+        gap = getproperty(dash_gap, modifier)
+        [dash, gap]
+    elseif linestyle == :dot
+        gap = getproperty(dot_gap, modifier)
+        [dot, gap]
+    elseif linestyle == :dashdot
+        gap = getproperty(dash_gap, modifier)
+        [dash, gap, dot, gap]
+    elseif linestyle == :dashdotdot
+        dash_gap = getproperty(dash_gap, modifier)
+        dot_gap  = getproperty(dot_gap,  modifier)
+        [dash, dash_gap, dot, dot_gap, dot, dash_gap]
     end
 end
 
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -677,11 +677,31 @@ convert_attribute(A::AbstractVector, ::key"linestyle") = A
 """
     A `Symbol` equal to `:dash`, `:dot`, `:dashdot`, `:dashdotdot`
 """
-function convert_attribute(ls::Symbol, ::key"linestyle")
-    return if ls == :solid
+convert_attribute(ls::Union{Symbol,AbstractString}, ::key"linestyle") = diff_pattern(ls, :normal)
+
+function convert_attribute(ls::Tuple{<:Union{Symbol,AbstractString},<:Any}, ::key"linestyle")
+    diff_pattern(ls[1], ls[2])
+end
+
+using UnPack
+
+function diff_pattern(linestyle, gaps)
+    float.([0.0; cumsum(pattern(linestyle, gaps))])
+end
+
+"The linestyle patterns are inspired by the LaTeX package tikZ as seen here https://tex.stackexchange.com/questions/45275/tikz-get-values-for-predefined-dash-patterns."
+
+function pattern(ls::Symbol, gaps = :normal)
+    if ls == :solid
         nothing
-    elseif ls in [:dash, :dot, :dashdot, :dashdotdot]
-        pattern(ls)
+    elseif ls == :dash
+        pattern("-", gaps)
+    elseif ls == :dot
+        pattern(".", gaps)
+    elseif ls == :dashdot
+        pattern("-.", gaps)
+    elseif ls == :dashdotdot
+        pattern("-..", gaps)
     else
         error(
             """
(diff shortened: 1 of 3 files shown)


src/conversions.jl
```diff
@@ -705,6 +694,32 @@ function convert_attribute(ls::Symbol, ::key"linestyle")
...
+end
+
+"The linestyle patterns are inspired by the LaTeX package tikZ as seen here https://tex.stackexchange.com/questions/45275/tikz-get-values-for-predefined-dash-patterns."
+function _pattern(linestyle, modifier=:normal;
+                  dot = 1,  dot_gap  = (normal = 2, dense = 1, loose = 4),
+                  dash = 3, dash_gap = (normal = 3, dense = 2, loose = 6)
+                  )
+
+    if linestyle == :dash
+        gap = getproperty(dash_gap, modifier)
```
jkrumbiegel, line 708: took me a bit to understand the logic here, could we simplify that somehow? also, how can anybody actually set the modifier? you just call `pattern(ls)`

jkrumbiegel:
ah also as a side note, you can leave out the `ax` parameter in the mutating functions if you plot into the last created axis :)

src/conversions.jl
greimel: Sorry, we could definitely hard-code the few patters that we have ;-)

But we could also use this function to allow more flexible patterns, that is any combination of dash and dot. And we could also make use of the modifiers. E.g. `("-....-.", :loose)`

jkrumbiegel: not necessarily hardcode, but I didn't see a path for a user to set the modifier. Also I like the idea with "-..-" etc, but you didn't add that yet as far as I can see?

greimel: You're right, it's not yet there.

jkrumbiegel: ok I was confused after you mentioned that on Slack :)

greimel:
Now, it is more flexible. Not sure if part of this would better fit in PlotUtils or so?

```
fig, ax, plt = lines(1:2, 2:3, linestyle="-....", linewidth=2)
lines!(1:2, 3:4, linestyle=(".", :dense), linewidth=2)
lines!(1:2, 4:5, linestyle=("-", :loose), linewidth=2)
lines!(1:2, 5:6, linestyle=(:dash, :dense), linewidth=2)
lines!(1:2, 6:7, linestyle=(:dash, 0.0), linewidth=2)
fig
```

image

greimel:
This is ready from my side.

CairoMakie works without changes.
GLMakie needs https://github.com/JuliaPlots/GLMakie.jl/pull/157.
WGLMakie doesn't support linestyle, so no action needed.

I am looking forward to your review.

src/conversions.jl
```diff
@@ -705,6 +712,61 @@ function convert_attribute(ls::Symbol, ::key"linestyle")
...
+        else
+            push!(pattern, dash_gap)
+        end
+    end
+    pattern
+end
+
+"Checks if the linestyle format provided as a string contains only dashes and dots"
+function check_pattern(ls_str)
+    for i in 1:length(ls_str)
```
jkrumbiegel, line 744: this would fail with unicode characters, because you can't index into strings like that. minor point, but maybe just replace with this small regex:

```julia
isnothing(match(r"^[\.\-]+$", ls_str)) && error...
```

src/conversions.jl
```diff
@@ -677,22 +677,29 @@ convert_attribute(A::AbstractVector, ::key"linestyle") = A
...
 """
-function convert_attribute(ls::Symbol, ::key"linestyle")
-    return if ls == :solid
+convert_attribute(ls::Union{Symbol,AbstractString}, ::key"linestyle") = pattern(ls, :normal)
+
+function convert_attribute(ls::Tuple{<:Union{Symbol,AbstractString},<:Any}, ::key"linestyle")
+    pattern(ls[1], ls[2])
+end
+
+function pattern(linestyle, gaps)
```
jkrumbiegel, line 686: can you call this something more obviously linestyle related? as names are visible across the whole codebase, I'd rather not have something so generic as `pattern`. Maybe `linestyle_pattern`, and for `diff_pattern` it could also be `linestyle_diff_pattern`

greimel: what about putting `pattern`, `diff_pattern`, `check_pattern`, `convert_gaps` in a module instead?

SimonDanisch: I'm for `line_pattern`, since they're 1D so it's mainly a line specific pattern.. The `style` seems a bit redundant in there.

src/conversions.jl
greimel: Thanks, I'll use your regex, even though I've no clue how to read it ;-)

src/conversions.jl
greimel: ok

src/conversions.jl
jkrumbiegel: it means:
`^` start of string
`[\.\-]` a set containing the . and - character
`+` at least one character from that set
`$` end of string

but I just checked and in this set . and - don't need to be escaped, you can just use `r"^[.-]+$"` then

test/unit_tests/conversions.jl
```diff
@@ -85,3 +85,38 @@ end
     @test ilabels == [1, 2]
     @test AbstractPlotting.categoric_position.(a, Ref(ilabels)) == [1, 1, 2]
 end
+
+import AbstractPlotting: check_pattern, diff_pattern
```
jkrumbiegel, line 89: imports are wrong here

greimel: thanks - updated.

src/conversions.jl
greimel: Thanks, TIL :-)

src/conversions.jl
```diff
@@ -705,6 +712,59 @@ function convert_attribute(ls::Symbol, ::key"linestyle")
...
+        end
+    end
+    pattern
+end
+
+"Checks if the linestyle format provided as a string contains only dashes and dots"
+function check_line_pattern(ls_str)
+    isnothing(match(r"^[.-]+$", ls_str)) &&
+        throw(ArgumentError("If you provide a string as linestyle, it must only consist of dashes (-) and dots (.)"))
+end
```
jkrumbiegel, line 746: ah yeah now you have to return nothing here to pass the tests, otherwise a Match will be returned

greimel: right, I should probably run tests locally before I push. -_-

greimel: ... but I didn't.

greimel:
tests pass again :-)

jkrumbiegel:
nice, I think this is good to go then. What do you think @SimonDanisch, is this a patch or minor release thing?

greimel:
it's technically breaking because it messes up patterns in GLMakie. Compat in GLMakie should only be bumped after the companion PR is merged there.

jkrumbiegel:
I don't quite understand why glmakie's behavior is different in the first place, doesn't the abstractplotting api specify someplace that we're using linewidth-dependent dash sizes? so that's what glmakie should have done anyway.

SimonDanisch:
I'm pretty sure GLMakie was intended to have linewidth scaled patterns... Not sure what's going on there, it's some pretty gnarly code.

greimel:
you are right. I was wrong. That's GLMakie before this PR.

image

So, what this PR does is changing the patterns in GLMakie. So, the dashing is more stretched.

greimel:
with this PR, but released GLMakie

image

not too bad

SimonDanisch:
So GLMakie actually doesn't need to change?

jkrumbiegel:
How is CairoMakie compared to that? It looks large

greimel:
As you wish. The open PR makes the looks more consistent. I'll post a picture in a sec.

jkrumbiegel:
Yeah I saw that you added a scaling factor there, but it's not good practice to add scaling factors by eye-balling :) There should be a correct version no?

greimel:
CairoMakie:

lines-Cairo

GLMakie:

lines-GL

(both with the current PR)

greimel:
> Yeah I saw that you added a scaling factor there, but it's not good practice to add scaling factors by eye-balling :) There should be a correct version no?

I agree, see my comment in the PR.

SimonDanisch:
Besides the weird artifact:
image
GLMakie seems to look more correct? (Sorry for asking seemingly rhetorical questions, but I'm just dropping in and haven't read through everything)

greimel:
What does more correct to you? If you mean the dash length, that's a matter of taste. The CairoMakie output approximately matches the pre-PR state of GLMakie.

jkrumbiegel:
I don't think GLMakie is better, the dash pattern should be 3 times the linewidth, that is what you see in the CairoMakie example. In the glmakie the small dashes are much too long I'd say, and the big dashes less so but still

SimonDanisch:
Ah I see.. The pattern looked visually more similar, but yeah you're right.

greimel:
@SimonDanisch compare tikZ line styles as given here: https://github.com/JuliaPlots/Makie.jl/issues/807#issue-781251725

jkrumbiegel:
Looking at this again, I think the multipliers we have here just don't look that good, even if cairomakie displays them correctly. In the tikz example the 3pt that we've translated into 3x multiplier is not a multiplier, and 3x is actually quite short for a dash as we can see above. So we should look at changing those multipliers to something more visually pleasing I guess?

greimel:
In my opinion, they look visually pleasing for linewidth > 2. Also linewidth = 2 looks nicer than linewidth = 1.

Here some patterns with linewidth = 2.

lines-Cairo

I would rather make the default linewidth == 2.

[see also the plot in https://github.com/JuliaPlots/AbstractPlotting.jl/pull/593#issuecomment-756290338 where linewidth = 2.]

jkrumbiegel:
huh yeah that looks better, maybe for small linewidths the fixed gaps actually look better, and only once the linewidth gets large enough it starts looking cramped. well I guess it's ok then :) we can also just go with it for a while and if people complain too much we make adjustments

PR #597, opened 2021-01-12
added a lock around cached_load().

norci:
for https://github.com/JuliaPlots/Makie.jl/issues/812

root cause:
if multiple threads call cached_load(), then push! will raise an error.


norci:
it still has problems. seems this commit does not work.

SimonDanisch:
Thanks for trying to take a stab at this... But Makie is right now not thread safe, and this PR doesn't really change that fact.
To actually make Makie threadsafe, we may need a different strategy than adding locks to places like this, so I'm going to close this PR for now.

PR #599, opened 2021-01-15
rename histogram to hist to be able to export it, add docs

jkrumbiegel:


mkborregaard:
Could `hist` be considered? That's the name for this function in MATLAB, Python (both Pandas and Matplotlib) and R. `histplot` on the other hand only exists in the seaborn package of python afaik and is less intuitive.

jkrumbiegel:
Sure, I have no strong opinion about this, just that unexported `histogram` is bad. If `hist` is more common across those ecosystems that's a good reason to choose the same name.

SimonDanisch (approved):

piever:
I also think the rename is a good idea! Otherwise both the type (`Histogram` vs `StatsBase.Histogram`) and the plotting function (`histogram` vs `AlgebraOfGraphics.histogram`) have conflicts. This way we are compatible with Python, where `numpy.histogram` is the computation and `matplotlib.pyplot.hist` is the plot. (Why shortened version would mean "plot" is a bit of a mystery to me, but as long as there is precedent I guess it's fine.)

jkrumbiegel:
I'm going to merge this now without an image example because the bug seems to be some kind of weird interplay between Documenter and StatsBase, and until that's fixed the example will just not have an image.

PR #610, opened 2021-01-21
Expanded convert function to take missing values

aramirezreyes:
Fixes https://github.com/JuliaPlots/Makie.jl/issues/819

I came accross this yesterday but this is my first PR at Makie so I would hope some guidance regarding if this is an appropriate fix. I can also try to throw some tests in tomorrow.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -260,7 +260,7 @@ outputs them in a Tuple.
 function convert_arguments(::SurfaceLike, x::AbstractVecOrMat{<: Number}, y::AbstractVecOrMat{<: Number}, z::AbstractMatrix{<: Union{Number, Colorant}})
     return (el32convert(x), el32convert(y), el32convert(z))
 end
-function convert_arguments(::SurfaceLike, x::AbstractVecOrMat{<: Number}, y::AbstractVecOrMat{<: Number}, z::AbstractMatrix{<:Number})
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Welcome and thanks! :) 
It looks like that change introduced ambiguities. Lets see if my change does the trick!
Some unit tests for this would be great!
Adding your example from the issue to this [file](https://github.com/JuliaPlots/AbstractPlotting.jl/blob/master/test/unit_tests/conversions.jl) in a new testset with some checks for the returned plot object should do the trick!

jkrumbiegel:
Can we also remove whatever we had previously that just converted unknown vectors, even something like `Vector{Any}` to a range 1:n? This causes some really weird bugs and should just error

jkrumbiegel:
I'm trying to fix this in a more general way here https://github.com/JuliaPlots/AbstractPlotting.jl/pull/619

My thinking is that for plotting, any missing in a float-like array can be replaced with NaN. So that code does this replacement for any argument if no convert_arguments method matches.

aramirezreyes:
I'm seeing that that solution should solve other problems apart from https://github.com/JuliaPlots/Makie.jl/issues/819, so it is fine if we close this one.

jkrumbiegel:
ok I'll close this then, thanks for getting the ball rolling again!

PR #611, opened 2021-01-21
allow passing a file path as a font

jkrumbiegel:


src/conversions.jl
```diff
@@ -795,6 +795,20 @@ function convert_attribute(x::Union{Symbol, String}, k::key"font")
     str = string(x)
     get!(_font_cache, str) do
         str == "default" && return to_font("Dejavu Sans")
+
+        # check if the string points to a font file and load that.
+        # let's assume fonts always have a dot extension so that normal font names
+        # would never accidentally be interpreted as files, because font names don't
+        # contain dots. this is pretty unlikely anyway.
+        if isfile(str) && occursin(".", basename(str))
```
SimonDanisch, line 803: shouldn't `isfile` be sufficient here? The behavior might be pretty frustrating, when someone loads a font without a dot...

SimonDanisch: I mean what are the odds of having a file lying around perfectly matching the font name, but the user doesn't actually want to load it? :D  Your dot rule should automatically apply there, if we assume no one has fonts on the disc, without dots in their name ;) Although, in that case we should guard against non fonts, so maybe not error when `try_load` returns nothing, but try to search instead?

jkrumbiegel: I don't know, yeah it would be weird :) maybe I got carried away thinking about the edge cases

PR #615, opened 2021-01-24
Add final code together at end of page

pbouffard:
For the impatient.

### DIFF
--- a/docs/src/makielayout/tutorial.md
+++ b/docs/src/makielayout/tutorial.md
@@ -572,10 +572,9 @@ ax1.ylabel = "Maximum Velocity [m/sec]"
 
 
 # Adding a Legend
-leg = fig[1, end+1] = Legend(fig,
+leg = fig[2, 1:2] = Legend(fig,
     [line1, scat1, line2, scat2],
     ["f(x) = x", "Data", "f(x) = -x + 26", "Data"])
-fig[2, 1:2] = leg
 
 
 # Fixing Spacing Issues
--- a/docs/src/makielayout/tutorial.md
+++ b/docs/src/makielayout/tutorial.md
@@ -532,3 +532,99 @@ documentation for more details and other dynamic parts like sliders and buttons!
 using GLMakie
 GLMakie.activate!()
 ```
+
+## Final code
+Here is the code producing the final figure above:
+```
+# Creating A Figure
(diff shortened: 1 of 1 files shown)


SimonDanisch (dismissed):

docs/src/makielayout/tutorial.md
```diff
@@ -532,3 +532,99 @@ documentation for more details and other dynamic parts like sliders and buttons!
 using GLMakie
 GLMakie.activate!()
 ```
+
+## Final code
+Here is the code producing the final figure above:
```
jkrumbiegel, line 537: Maybe add:

"Note that this code is not perfect, but simply reflects the tutorial content including unnecessary intermediate steps like misplacing and then moving the legend."

jkrumbiegel (changes requested):

docs/src/makielayout/tutorial.md
pbouffard: Not sure what you mean by the unnecessary intermediate steps, can you elaborate?

jkrumbiegel: Well I deliberately place the legend in the "wrong" place first to show that you can move it, but of course if you run the whole script at once it's a bit silly to include that. But it's not supposed to be streamlined code, just the collection of all steps

jkrumbiegel: now you removed that step, but I wouldn't want the code down there to be different. just tell people that it's not how the "ideal" code would be written. it's the same with the heatmap example, they are also moved after creation.

jkrumbiegel: it's a kind of tutorial where some less than ideal things are shown first and then resolved

SimonDanisch:
@jkrumbiegel I guess we only need a short notice before the code like:
```
Note that this code is just copy & pasted from the tutorial steps, and not representative of how you'd write it in one go.
```
To merge this?

jkrumbiegel:
ok I had thought about it a bit more, and I've added something similar with the updated code to the beginning of the page, but in a detail view, so click to expand. I think that's visible but also doesn't get in the way. https://github.com/JuliaPlots/AbstractPlotting.jl/pull/745 I'll close this PR then, thanks for sticking with it

PR #616, opened 2021-01-25
Document colgap! and rowgap!

Kolaru:
Add mention of `colgap!` and `rowgap!` in the documentation with simple examples.

### DIFF
--- a/docs/src/makielayout/grids.md
+++ b/docs/src/makielayout/grids.md
@@ -294,19 +294,19 @@ columns respectively.
 ```@example spacing
 using CairoMakie
 
-scene, layout = layoutscene(resolution = (1200, 900))
+fig = Figure(resolution = (1200, 900))
 
-axs = layout[] = [Axis(scene) for i in 1:3, j in 1:3]
+axs = [Axis(fig[i, j]) for i in 1:3, j in 1:3]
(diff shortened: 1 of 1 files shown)


docs/src/makielayout/grids.md
```diff
@@ -286,6 +286,43 @@ save("example_trimming_3.svg", scene); nothing # hide
...
+
+You can use `rowgap!` and `colgap!` to change the spacing between rows or
+columns respectively.
+
+```@example spacing
+using CairoMakie
+
+scene, layout = layoutscene(resolution = (1200, 900))
+
+axs = layout[] = [Axis(scene) for i in 1:3, j in 1:3]
```
jkrumbiegel, line 299: nice, thank you! but could you switch the example to the `Figure` workflow? I assume people will mostly try to change figures and then they can see that you have to access `fig.layout` to call layout specific functions on it.

the axs part could then also be the new syntax `axs = [Axis(f[i, j]) for i in 1:3, j in 1:3]`

jkrumbiegel (changes requested):

docs/src/makielayout/grids.md
Kolaru: Done.

jkrumbiegel (approved):

PR #617, opened 2021-01-25
jk/categorical colormap and contourf

jkrumbiegel:
- colorbar shows a number of rects instead of heatmap for irregular maps
- store contourf levels separately and use in colorbar

### DIFF
--- a/docs/src/makielayout/layoutables_examples.md
+++ b/docs/src/makielayout/layoutables_examples.md
@@ -13,6 +13,77 @@ Pages = ["layoutables_examples.md"]
 Depth = 2
 ```
 
+## Colorbar
+
+A Colorbar needs a colormap and a tuple of low/high limits.
+The colormap's axis will then span from low to high along the visual representation of the colormap.
+You can set ticks in a similar way to `Axis`.
+
+Here's how you can create Colorbars manually.
+
+```@example
+using CairoMakie
+
+fig = Figure(resolution = (1200, 900))
+
+Axis(fig[1, 1])
+
+# vertical colorbars
+Colorbar(fig[1, 2], width = 25, limits = (0, 10), colormap = :viridis,
+    flipaxisposition = false, ticklabelalign = (:right, :center))
+Colorbar(fig[1, 3], width = 25, limits = (0, 5),
+colormap = cgrad(:Spectral, 5, categorical = true))
+Colorbar(fig[1, 4], width = 25, limits = (-1, 1), colormap = :heat,
+    highclip = :cyan, lowclip = :red, label = "Temperature")
+
+# horizontal colorbars
+Colorbar(fig[2, 1], height = 25, limits = (0, 10), colormap = :viridis,
+    vertical = false, ticklabelalign = (:center, :bottom))
+Colorbar(fig[3, 1], height = 25, limits = (0, 5),
+    colormap = cgrad(:Spectral, 5, categorical = true), vertical = false,
+    ticklabelalign = (:center, :bottom))
+Colorbar(fig[4, 1], height = 25, limits = (-1, 1), colormap = :heat,
+    label = "Temperature", vertical = false, flipaxisposition = false,
+    ticklabelalign = (:center, :top), highclip = :cyan, lowclip = :red)
+
+fig
+
+save("example_colorbar.svg", fig); nothing # hide
+```
+
+![example colorbar](example_colorbar.svg)
+
+You can also automatically choose colormap and limits for certain plot objects by passing them as the second argument.
+
+```@example
+using CairoMakie
+
+xs = LinRange(0, 20, 50)
+ys = LinRange(0, 15, 50)
+zs = [cos(x) * sin(y) for x in xs, y in ys]
+
+fig = Figure(resolution = (1200, 900))
+
+ax, hm = heatmap(fig[1, 1][1, 1], xs, ys, zs)
+Colorbar(fig[1, 1][1, 2], hm, width = 20)
+
+ax, hm = heatmap(fig[1, 2][1, 1], xs, ys, zs, colormap = :heat)
+Colorbar(fig[1, 2][1, 2], hm, width = 20)
+
+ax, hm = heatmap(fig[2, 1][1, 1], xs, ys, zs, colormap = :grays,
+    colorrange = (-0.75, 0.75), highclip = :red, lowclip = :blue)
+Colorbar(fig[2, 1][1, 2], hm, width = 20)
+
+ax, hm = contourf(fig[2, 2][1, 1], xs, ys, zs,
+    levels = -1:0.25:1)
+Colorbar(fig[2, 2][1, 2], hm, width = 20, ticks = -1:0.25:1)
+
+fig
+save("example_colorbar_2.svg", fig); nothing # hide
+```
+
+![example colorbar 2](example_colorbar_2.svg)
+
 ## Slider
 
 A simple slider without a label. You can create a label using a `Label` object,
--- a/docs/src/makielayout/layoutables_examples.md
+++ b/docs/src/makielayout/layoutables_examples.md
@@ -67,15 +67,16 @@ fig = Figure(resolution = (1200, 900))
 ax, hm = heatmap(fig[1, 1][1, 1], xs, ys, zs)
 Colorbar(fig[1, 1][1, 2], hm, width = 20)
 
-ax, hm = heatmap(fig[1, 2][1, 1], xs, ys, zs, colormap = :heat)
+ax, hm = heatmap(fig[1, 2][1, 1], xs, ys, zs, colormap = :grays,
+    colorrange = (-0.75, 0.75), highclip = :red, lowclip = :blue)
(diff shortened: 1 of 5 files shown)


jkrumbiegel:
with an MWE from @briochemc:

```julia
using AbstractPlotting.KernelDensity
using Random
Random.seed!(0)
# Fake data
N = 1000
x = 2randn(N) .+ 1
y = sin.(x) + randn(N)
# density
D = kde((x, y))
# quantiles of density
δx = step(D.x)
δy = step(D.y)
Q = vec(D.density) * δx * δy
idx = sortperm(Q)
Q_sorted = Q[idx]
Dcum = similar(D.density)
Dcum[idx] .= 1 .- cumsum(Q_sorted)
# plot
fig = Figure(resolution = (600, 500))
levels = 0.1:0.1:0.9

ax, co = contourf(fig[1,1], D.x, D.y, Dcum; levels)

Colorbar(fig[1, 2], co, width = 25)

fig
```
grafik

and with `levels = [0.1, 0.3, 0.4, 0.7, 0.9]` irregular colorbars also work

grafik

jkrumbiegel:
The implementation is a bit "weird" and there are probably corner cases, although the typical use case should be fine.

Basically the colorbar internally wants to have either a `ContinousColorGradient` or a `CategoricalColorGradient` and `limits` from low to high. If we have a `ContinousColorGradient` then the `nsteps` attribute controls the resolution of the colorbar. If we have a `CategoricalColorGradient` instead, the number of boxes is determined by looking at the `values` vector of the gradient.

The special method for making a Colorbar from a `contourf` then looks at the `_computed_levels` attribute and the colormap and creates `limits` and a fitting `CategoricalColorGradient` from them.

It doesn't feel super clean, yet, but at least I like that we gain the ability to have irregular categorical colormaps.

mkborregaard:
so quick, and looks amazing. Question - should the colorbar ticks follow the cutoff levels?

jkrumbiegel:
they could, but for a lot of data were you just divide into `n` levels automatically, the numbers will not be nice at all. I think this will probably be left to the user, although one could make some shortcut available for creating a categorical colorbar where you want to label the cells and not the gaps

mkborregaard:
Unless you'd use a histogram-like method to create the color levels - in a histogram with `n` bins, the bins align along "nice" numbers - such as multiples of 2, 5 etc as integers. It would make sense to do something similar for n arbitrary color bins https://github.com/JuliaStats/StatsBase.jl/blob/0ffe2e20998d98eb514fa0b08090cf05a18fa7b5/src/hist.jl#L26-L105

jkrumbiegel:
Interesting, I never noticed that. We can think about adding that here, too

mkborregaard:
Yeah StatsBase is already a dep so I guess this could just call the `histrange` function and be done with it.

jkrumbiegel:
added the triangles here for good measure 
grafik

briochemc:
This is amazing! 

---

Question: Is there an issue with adding the `highclip` and `lowclip` kwargs to `contourf`?

---

Request: I would also like to suggest a different default behavior, if that's not too much. I really like the `extend` keyword because that's the behavior I would be expecting. The reason is that, unlike in the images I showed, I think most of the time, one wants one (or both) extensions to be the extreme colors of the colormap. For example, I would expect
```julia
ax, co = contourf(fig[1,1], x, y, z, levels=0:1:10, extend=:both, colormap=cmap)
cbar = fig[2,1] = colorbar(fig, co)
```
to do the following:
- in the contour plot, color everything beyond the levels according to the colormap `cmap`. I.e., be equivalent I think to 
    ```julia
    ax, co = contourf(fig[1,1], x, y, z, levels=[-Inf; 0:1:10; Inf], colormap=cmap)
    ```
- in the colorbar, have rectangles for all values within `levels=0:1:10` colored with the `n-2` "inside" colors of the corresponding categorical `cmap`, and triangles colored with the 1st and last colors.


If, instead, one wanted to specify the color of one of the extension, e.g., with something like one of those (not sure what you'd prefer for an API)

```julia
ax, co = contourf(fig[1,1], x, y, z, levels=0:1:10, extend=(:white, :auto), colormap=cmap)
ax, co = contourf(fig[1,1], x, y, z, levels=0:1:10, extendlow=:white, extendhigh=:auto, colormap=cmap)
```
which could be equivalent to 
```julia
ax, co = contourf(fig[1,1], x, y, z, levels=[0:1:10; Inf], colormap=cmap)
```
but then with a colorbar with rectangles filled with the `n-1` first colors of `cmap`, with the high-end triangle filled with the last color, and with the low-end triangle in `:white`.

I apologize if that's too much to ask, I really appreciate all the amazing work you've put in everytime I've asked for something @jkrumbiegel ! Thank you so much for all of that!

jkrumbiegel:
with `extendlow = :magenta, extendhigh = :cyan` 
grafik

jkrumbiegel:
and automatic
grafik

briochemc:
You're amazing @jkrumbiegel!

lazarusA:
> and automatic
> grafik

Wow!!! this is going to fast...

mkborregaard:
Does my eye deceive me, or is the tickmark just below the transition between swatches?

jkrumbiegel:
haha I've seen this as well. I'm not 100% sure what the reason is right now. Could be:
- antialiasing artifact from Cairo
- imperfect sub-pixel drawing from Cairo
- swatch boundary is simply not exactly 0.6 and the image is correct, the tick is just confusing (would have to check the iterable again)
- something wrong with the tick code
- something wrong with the colorbar code

mkborregaard:
Another possibility I considered is that there's a small boundary line between the swatches, which is lighter, and thus perceived as belonging to the lower swatch? Anecdotally, here they all seem off in the direction of the lighter swatch.
Skærmbillede 2021-01-26 kl  16 29 37

mkborregaard:
The time from Slack conversation to closed issue and merged awesome PR is just amazing here @jkrumbiegel . This is the kind of thing that is putting Makie on the path to world dominance :-)

briochemc:
Just thought I'd comment here to not pollute other issues and hear what you have to say about it before submitting yet another issue.

I just noticed the fine boundaries of the colorbar that Michael mentioned, and at the same time I don't see these in `contourf` plots. I guess this means they (`contourf` and `Colorbar`) don't use the same primitive and I'm sure that there was a reason for that but I'm not sure what that reason is?

I bring this up because I think of colorbars as the same thing as the plot they refer to, except that the colorbar's data is just a range. For example, say a plot was a combination of a heatmap and some contour lines, something like

```julia
heatmap!(ax, x, y, z, colorrange=extrema(levels))
contour!(ax, x, y, z, levels=levels)
```

then the corresponding colorbar could be built with calls similar to 

```julia
heatmap!(cbar_ax, [0,1], levels, colorrange=extrema(levels))
contour!(cbar_ax, [0,1], levels, levels=levels)
```

of which all the arguments can be grabbed from `ax` AFAIU. (the `[0,1]` is just a dummy interval for the direction perpendicular to the color range.)

Similarly, a plot with filled contours at levels `levels1` and some specific contour lines at different levels (`levels2`), like

```julia
contourf!(ax, x, y, z, levels=levels1)
contour!(ax, x, y, z, levels=levels2)
```

could then provide all the arguments to make a matching colorbar with something like

```julia
contourf!(cbar_ax, [0,1], levels1, levels=levels1)
contour!(cbar_ax, [0,1], levels2, levels=levels2)
```

Just thought I should write this somewhere in case it's useful.

briochemc:
Just thought I'd add a picture to explain what I mean:

Screen Shot 2021-02-01 at 6 00 39 pm

In this screenshot (made with Makie) I plot, on top of each other:
- a `contourf` (and matching colorbar, thanks to this amazing PR! 🎉 )
- a few `contour` lines (but not at all levels)
- a `scatter` plot

It would be great to somehow be able to "add" the `contour` lines to the colorbar. 

Right now, if I do
```
contourf!(ax, x, y, z, levels=levels1)
contour!(ax, x, y, z, levels=levels2)
```
then `ax` contains both the filled contours and the contour lines:
```julia
julia> ax.plots
2-element Array{AbstractPlot,1}:
 Combined{AbstractPlotting.contourf,Tuple{Array{Float32,1},Array{Float32,1},Array{Float32,2}}}
 Combined{AbstractPlotting.contour,Tuple{Array{Float32,1},Array{Float32,1},Array{Float32,2}}}
```

but I can't feed `ax.plots` directly to the colorbar. Maybe the `Colorbar` function could be extended to accept `Vector{AbstractPlot}`? (Not sure what would be the way to implement this.)

FWIW, a matplotlib example of just that is available in [their `contourf` demos](https://matplotlib.org/3.3.3/gallery/images_contours_and_fields/contourf_demo.html#sphx-glr-gallery-images-contours-and-fields-contourf-demo-py):

briochemc:
A last comment I would like to make about the automatic setting:



It would maybe be a bit nicer to have the categorical colormap default be of size `n+2` where `n` is the number of colors inside (within) the colorbars' extrema, so that the color of the triangles are different from the rest.

PR #618, opened 2021-01-27
Fix admonitions

sostock:
A `warning` and a `tip` in the docs are not printed correctly. This PR fixes that.


jkrumbiegel:
thanks!

PR #619, opened 2021-01-27
jk/conversion fixes

jkrumbiegel:
- remove blanket categorical conversion
- remove another generic categorical convert
- add convert_single_argument and a conversion for missing reals


piever:
Nice!

I wanted to ask, what is the situation for plotting categorical data with this PR? Things like `barplot(["a", "b", "c"], rand(3))`, or with `CategoricalValue` from CategoricalArrays instead of `String`s?

jkrumbiegel:
The problem was that the way it worked before treated any unknown vector type as categorical, which is kind of overdoing it. Right now we don't have any categorical shortcuts implemented. I would like to re-add them, but I'm still wondering how to best do so. For most cases it would probably be fine to just do the tick relabeling "trick" without any further work, but I'm doubtful that this doesn't lead to difficult problems when using other mutating plotting functions which don't "know" about the first categories that the axis was initialized with.

This is actually much easier for AlgebraOfGraphics to solve, because you know the categories present in all parts of the plot, and can compute an appropriate axis for the union of them. It's the same in ggplot. I'm not sure what the best way is for Makie.

piever:
Yes, here it's kind of problematic, see for example https://github.com/JuliaPlots/Makie.jl/issues/453. I'm also not sure what would be a good fix.

jkrumbiegel:
I think it's better to rip out the existing signatures for now which actually don't really work and cause more trouble than they're worth, while we're trying to come up with something better.

mkborregaard:
Would it not work to just update the observable controlling the axis when adding new series?

piever:
This is a bit more complex than expected to rebase correctly, as some tests were added that rely on the old behavior. Continuing in #737

PR #624, opened 2021-02-03
fix ssao rendering issue in LScene and update lighting docs

ffreyer:
Docs update for https://github.com/JuliaPlots/GLMakie.jl/pull/107.

Also fixes this issue with ssao in figures:
Screenshot from 2021-02-03 10-33-40

Currently SSAO distinguishes between scenes using a stencil buffer that is only written to if `scene.clear = true`. So I changed the default for `LScene`. This means an `LScene` will not render over other plots that are rendered before.

I tried a bunch of variations of 
```
fig = Figure()
Menu(fig[1, 1])
s = LScene(fig[2, 1])
scatter!(s, rand(10))
```
including 3d plots with ssao, 2d without, 2d without fxaa and different orders of operations (Menu-scene-plot, scene-menu-plot, scene-plot-menu). The Menu is always on top when opened, which is what I thought might break. 

This also means that `backgroundcolor` now works for `LScene`.

### DIFF
--- a/docs/Project.toml
+++ b/docs/Project.toml
@@ -14,3 +14,4 @@ Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 
 [compat]
 julia = "1.3"
+GLMakie = "0.1.26"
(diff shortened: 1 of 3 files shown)


ffreyer:
GLMakie is one version behind in the docs build...?

SimonDanisch:
You should be able to bump the version in AbstractPlotting\docs\Project.toml

SimonDanisch:
hm...

PR #626, opened 2021-02-03
Ffreyer docs

SimonDanisch:


SimonDanisch:
@ffreyer, non commiters don't get a preview :( 
But here you are: http://makie.juliaplots.org/previews/PR626/lighting.html
Is this really with SSAO?
image

ffreyer:
No :(

This is how it looks when I run it manually:
Screenshot from 2021-02-04 00-29-21

ffreyer:
`GLMakie.enable_SSAO[] = true` requires shaders to recompile, maybe that's causing an issue?

SimonDanisch:
that could be...maybe we need to close(current_screen()) somewhere

SimonDanisch:
Alright, that worked... can you check if this looks good to you? http://makie.juliaplots.org/previews/PR626/lighting.html
And @jkrumbiegel I saw you discussed the changes introduced here on slack and am not sure what the conclusion was...Could you give your approval?

ffreyer:
The ssao plot looks good now.

I did a few tests with the change like I described in the original pr, they all looked normal. Maybe there are some edge cases where this changes something, but then you can just use `LScene(scenekw = (clear=false, ))` to get the old behavior back. I think it's fine.

jkrumbiegel (approved):
I think it's fine, we'll see if the clearing trips a lot of people up when other things overlap, I guess

PR #629, opened 2021-02-07
remove unnecessary backticks from docs

simonbyrne:
See http://makie.juliaplots.org/stable/scenes.html#AbstractPlotting.update_cam!


SimonDanisch:
Thanks!

PR #632, opened 2021-02-07
Improve error message for missing method

simonbyrne:
Ideally it would also print out a list of applicable methods, but in the meantime this sort of works.


SimonDanisch:
Thank you, that looks better!

PR #634, opened 2021-02-08
events refactor - PriorityObservable

ffreyer:
Attempt number two of  #586.
Should fix https://github.com/JuliaPlots/Makie.jl/issues/777 and probably some others...
Requires https://github.com/JuliaPlots/GLMakie.jl/pull/154 and https://github.com/JuliaPlots/WGLMakie.jl/pull/89

This adds a `PriorityObservable` which mainly does two things:
- It goes through connected functions in order of priority. There can be multiple functions with the same priority - these should be considered as executed at the same time.
- Every observer function must return a `Bool`. If it returns `true` triggering of any other observer functions is cancelled. This means that `map` doesn't work.

This type of observable is used for the event system to allow observer functions to be called in a (partially) ordered manner, and to allow one observer function to disable another.

For testing:

```julia
fig = Figure()
on(println, Toggle(fig[1, 1], tellwidth = false, tellheight = false).active)
on(println, Textbox(fig[1, 2], tellwidth = false, tellheight = false).stored_string) # NO
on(println, Menu(fig[1, 3], tellwidth = false, tellheight = false).selection) # OK
on(println, Slider(fig[2, 3], tellwidth = false, tellheight = false).value) # NO
on(println, IntervalSlider(fig[3, 3], tellwidth = false, tellheight = false).selected_indices) # NO
on(println, Button(fig[4, 3], tellwidth = false, tellheight = false).clicks) # OK
s = LScene(fig[2:4, 2])
a = Axis(fig[2:4, 1], tellwidth = false, tellheight = false)
scatter!(a, rand(Point2f0, 10))
scatter!(s, rand(Point3f0, 10))
mesh!(s, Sphere(Point3f0(0), 1f0))
fig
```


ffreyer:
Basic functionality is all up and running. Most of what's left to do is to make proper use of it and clean up things.

Quick TODO:
- [ ] Work out how MakieLayout interactions should work with this.
- [ ] Make use of the extra information in `mousebutton` to simplify mousestatemachine (This will probably mean some minor changes to behavior)
- [ ] Allow mousestatemachine to use boundingboxes instead of plots for better performance. (I noticed that was the main thing costing performance when testing large `Toggle`-grids.)
- [ ] Maybe a better way to error on observer function return types?
- [ ] Maybe a better way to get whether a `PriorityObservable` had any consuming observer functions. (Currently `setindex!` returns that, which might be a bit weird and doesn't work with `consumed = obs[] = value` outside the global scope...)

ffreyer:
I updated the MouseStateMachine implementation to work with bounding boxes and changed every interactive layoutable to use that (except the ones that use scene) Timings with TimerOutputs.jl for the toggle-grid above are now:

```
 ────────────────────────────────────────────────────────────────────────────
                                     Time                   Allocations      
                             ──────────────────────   ───────────────────────
      Tot / % measured:           30.2s / 0.80%           1.06GiB / 1.01%    

 Section             ncalls     time   %tot     avg     alloc   %tot      avg
 ────────────────────────────────────────────────────────────────────────────
 mouseposition          549   78.8ms  32.8%   144μs   5.85MiB  53.7%  10.9KiB
 correct_mouse          634   77.4ms  32.2%   122μs     0.00B  0.00%    0.00B
 GLFW.GetCursorPos      634   51.2ms  21.3%  80.8μs     0.00B  0.00%    0.00B
 mousebutton            176   32.8ms  13.6%   186μs   5.04MiB  46.3%  29.3KiB
 window_open              1   13.9μs  0.01%  13.9μs      144B  0.00%     144B
 window_area              1   12.1μs  0.01%  12.1μs      432B  0.00%     432B
 window_dpi               1   5.28μs  0.00%  5.28μs      128B  0.00%     128B
 entered_window           2   2.02μs  0.00%  1.01μs      160B  0.00%    80.0B
 ────────────────────────────────────────────────────────────────────────────
```
where each event timing is a `@timeit event[] = new_value`. I also wanted to highlight that `correct_mouse` is pretty expensive and apparently `GLFW.GetCursorPos` is too. Maybe we can figure out something better there...

ffreyer:
This should be very close to done, except for maybe some more cleanup/tests/docs/performance tweaks.

src/interaction/PriorityObservable.jl
```diff
@@ -0,0 +1,142 @@
...
+        sanitized_func = x -> begin f(x); false end
+        @warn(
+            "Observer functions of PriorityObservables must return a Bool to " *
+            "specify whether the update is consumed (true) or should " *
+            "propagate (false) to other observer functions. The given " *
+            "function has been wrapped to always return false."
+        )
+    else
+        sanitized_func = f
+    end
```
ffreyer, line 76: This should allow the old syntax to work, but with a warning

src/interaction/iodevices.jl
```diff
@@ -138,8 +138,17 @@ module Keyboard
...
+    
+    Enumerates all key states/actions in accordance with the GLFW spec.
 
+    $(INSTANCES)
+    """
+    @enum Action begin
+        release = 0
+        press   = 1
+        repeat  = 2
+    end
```
ffreyer, line 151: In the last pr I tried to differentiate "is being pushed", "is held", "is being released" and "is up". I ended up with "press" and "release" but "down" and "up" are probably more common. Should I change this to down/up? Or maybe add them as aliases?

src/types.jl
```diff
@@ -14,82 +14,152 @@ end
...
     """
-    mousebuttons::Node{Set{Mouse.Button}}
+    mousebutton::PriorityObservable{MouseButtonEvent}
     """
-    The position of the mouse as a `NTuple{2, Float64}`.
-    Updates whenever the mouse moves.
+    A Set of all currently pressed mousebuttons.
     """
-    mouseposition::Node{NTuple{2, Float64}}
+    mousebuttonstate::Node{Set{Mouse.Button}}
```
ffreyer, line 65: `mousebuttonstate` and `keyboardstate` are basically what `mousebuttons` and `keyboardbuttons` are on master, though `keyboardstate` doesn't trigger on repeats. They're still useful to have (e.g. for `ispressed`) but I'm not sure in what form.

For compat it would be nice to have them as `mousebuttons` and `keyboardbuttons` nodes, doing the same thing they used to.

Conceptually in the new design they shouldn't be nodes - you should react to `keyboardbutton` and `mousebutton` instead. They should also be easy to differentiate from those names.

So what should we do here? We could also do both and have the old `mousebuttons` and `keyboardbuttons` around for deprecation purposes while implementing buttonstates as sets.

ffreyer: I changed states to be just `Set`s and added a `getproperty` that generates nodes for `mousebuttons` and `keyboardbuttons` which should behave like on master.

src/interaction/PriorityObservable.jl
ffreyer: I also added a `show_backtrace(...)` here now, so it's easier to find what's causing the warning. Is it possible to attach that to a warning directly?

ffreyer:
I re-added `DragEnum` and added some compatibility code for `events.mousebuttons`, `events.keyboardbuttons` and `events.mousedrag`.  I think now everything that used to work but got removed is available again, with deprecation warnings.

ffreyer:
May need changes when #642 gets merged

ffreyer:
I want to re-implement `PriorityObservables` similar to normal Observables in 0.4. I'll keep this as a draft until I do.

ffreyer:
Back to draft till I make adjustments for #675 and possibly #674

ffreyer:
Should be fine again

ffreyer:
Needs updates for Axis3

ffreyer:
Needs updates for https://github.com/JuliaPlots/AbstractPlotting.jl/pull/691 as well (mouseevents at least)

src/camera/camera.jl
```diff
@@ -12,8 +12,17 @@ end
 
 function disconnect!(c::Camera)
     for node in c.steering_nodes
-        filter!(listeners(node)) do x
-            !(x isa CameraLift) # remove all camera lifts
+        # remove all camera lifts
+        if node isa PriorityObservable
+            for f in listeners(node)
```
SimonDanisch, line 17: Should we use the same code here for normal observables?

ffreyer: Good point

SimonDanisch:
Do the tests need fixing, or do we just need to increase the accepted error?

ffreyer:
Oh, I kinda assumed tests would be failing here because the GLMakie and WGLMakie prs are needed for things to work. Let me check

ffreyer:
Those are tests I added. I'm not sure why they're failing now, but I think we can just update the values. I'll do that and make another commit. If those break frequently we can also remove them again, I guess

SimonDanisch:
Interesting, that this doesn't seem to be deterministic...

ffreyer:
The numbers in 527af0 came from running the test code in the terminal. That didn't work for the tests locally either, but setting a fixed resolution did fix it locally. Let's see...

ffreyer:
I checked the first example that failed in the docs, it works locally. So those should just be because GLMakie is behind, I think. I don't know why `record` cares about events though...

SimonDanisch:
Thanks a lot, that's a great addition! :)

PR #637, opened 2021-02-10
add PlotMethodError, suggest alternatives

simonbyrne:
I think this gives slightly more useful error messages:

Screen Shot 2021-02-10 at 1 40 14 PM


SimonDanisch:
Thanks, that's a good idea! :)

PR #650, opened 2021-03-05
Loosen signature on FileIO.save specialization

timholy:
Fixes https://github.com/JuliaPlots/Makie.jl/issues/864

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "AbstractPlotting"
 uuid = "537997a7-5e4e-5d89-9595-2241ea00577e"
 authors = ["Simon Danisch", "Julius Krumbiegel"]
-version = "0.15.23"
+version = "0.15.24"
 
 [deps]
 Animations = "27a7e980-b3e6-11e9-2bcd-0b925532e340"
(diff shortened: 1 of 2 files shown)


timholy:
So basically the issue is that you need this method to have priority over `save(file, args...; options...) in FileIO at /home/tim/.julia/packages/FileIO/ajK7j/src/loadsave.jl:125` even when the first argument is just a file name. Since the `fig` is a specialization over `args...`, that's all that should be required.

The old version worked by being called by FileIO internals. Now this is the method that gets called directly by the caller.

SimonDanisch:
Can you rebase master?
This commit doesn't seem to be included: https://github.com/JuliaPlots/AbstractPlotting.jl/commit/b1a124b0d4e52af5ef393dc060d208c204ca3b3d

timholy:
https://github.com/timholy/AbstractPlotting.jl/commits/teh/fileiosave

Seems like it's there.

FYI I ran this also with FileIO1.5 & it passes there too. I didn't try any earlier versions, but I'm guessing they'd all be the same.

SimonDanisch:
Ah, maybe I only tested that commit on julia 1.6, and 1.4 works differently for name clashes...

timholy:
I think the errors are all due to the `volume` issue. Worth merging anyway?

SimonDanisch:
Thank you! I wanted to make tests pass first, and then a busy weekend came ;)

PR #657, opened 2021-03-08
use mesh for 3D arrows

ffreyer:
## Changes this
Screenshot from 2021-03-08 20-51-08

## into this
~~https://user-images.githubusercontent.com/10947937/110376337-5b757f80-8053-11eb-85b0-139926e8f410.png~~
~~https://user-images.githubusercontent.com/10947937/110374315-d721fd00-8050-11eb-825b-45d96847ed8b.png~~
~~https://user-images.githubusercontent.com/10947937/110397892-fd589480-8072-11eb-8822-8ed5df4c7728.png~~
Screenshot from 2021-03-10 15-49-53

which looks way better imo.


mschauer:
Are there performance considerations to be taken when plotting a large number of arrows?

ffreyer:
The mesh is more complex so yes, there is. I can plot 100 000 arrows with a little bit of lag (I have a 6GB 1060), 1 000 000 is still workable. The old version performs similarly with 10x more arrows.

However if you desperately need to plot 10 000 000 arrows you can set `arrowhead = Pyramid(Point3f0(0, 0, -0.5), 1f0, 1f0)` and get the old plot without the lines.

I updated arrows to shade smoothly. Maybe they're to thin now?
Screenshot from 2021-03-09 00-49-15

And for the sake of it, here are 1 000 000 of them
Screenshot from 2021-03-09 01-00-33

SimonDanisch:
It does look much nicer :)

ffreyer:
TODO:
- [ ] separate mesh into arrow head and tail parts
- [ ] recover attributes/controls (e.g. dir -> arrow scale, lengthscale, ...)

ffreyer:
I added an `align` attribute but I'm not sure how to communicate the options well. I added two modes:
* align the start of the tail (line/cylinder) to the given positions. I'm thinking of this as `align = :origin` or `:tail` but it's currently the else-branch so anything goes. It's also the default and the old behavior.
* align the connection point between the tail and the head (line/cylinder to arrow head) to the given positions. I thought of this as `align = :center` before, but that doesn't make sense if you can independently scale the tail and the head. I have it as `:head` now which might be confusing?

I'm also wondering if I should rename `line...` attributes to `tail...`.

ffreyer:
I moved the scaling options into `plot!` so they can be different for for 2d and 3d. Now `lengthscale =1` and `arrowsize = 1` mean that both the tail and the tip have length 1 or the length of the given direction (depending on normalize). `linewidth = 1` and `arrowsize = 1` result in a diameter 1 tube and cone. 

The 3d defaults are `linewidth = 0.05`, `lengthscale = 0.3` and `arrowsize = Vec3f0(0.2, 0.2, 0.3)` which is similar to the previous version. 2d should be the same.
Screenshot from 2021-03-10 15-49-53

ffreyer:
TODO
- [ ] generate arrow components instead of using obj files

ffreyer:
This also changes the marker in 3D `streamplot`, though not drastically. It's just a cone now instead of a square pyramid. I added `arrow_head` and `quality` there too, so you can adjust those if you want.

src/basic_recipes/arrows.jl
```diff
@@ -24,15 +24,16 @@ $(ATTRIBUTES)
...
         arrowhead = automatic,
-        arrowtail = nothing,
+        arrowtail = automatic,
         linecolor = :black,
-        linewidth = 1,
-        arrowsize = 0.3,
+        linewidth = automatic,
+        arrowsize = automatic,
         linestyle = nothing,
-        scale = Vec3f0(1),
```
ffreyer, line 32: Is it ok to remove scale?

ffreyer:
TODO:
- [ ] fix docs for changes made in #659

ffreyer:
Should we merge this now?

ffreyer:
I added a bunch more default attributes which were previously not passed. To avoid issues like https://discourse.julialang.org/t/makie-arrow-through-transparent-mesh/59666

SimonDanisch:
merged in #731

PR #662, opened 2021-03-18
clearer error messages and bugfix for nested figuresubpositions

jkrumbiegel:


ericphanson:
might be good to have a `@test Axis(Figure()[1,1][1,1][1,1]) isa Axis` test or something like that, to address the MWE

SimonDanisch (approved):

PR #665, opened 2021-03-20
update record docstring to figures

ffreyer:


src/display.jl
```diff
@@ -496,14 +496,15 @@ function save(path::String, io::VideoStream;
...
+    record(func, figure, path, iter;
             framerate = 24, compression = 20, sleep = true)
 
-The first signature provides `func` with a VideoStream, which it should call `recordframe!(io)` on when recording a frame.
+The first signature provides `func` with a VideoStream, which it should call 
+`recordframe!(io)` on when recording a frame.
 
-Records the Scene `scene` after the application of `func` on it for each element
-in `itr` (any iterator).  `func` must accept an element of `itr`.
+Records the Figure (or Scene) `figure` after the application of `func` on it for
```
jkrumbiegel, line 506: application of func *on it* is kind of misleading, as it's f(itr_element) not f(figure).

Maybe: "Records the Figure (or Scene) `figure`, where a frame is saved after each execution of `func(iter_element)` for all elements in `iter`."

ffreyer:
The docs failure isn't related to my changes, is it?

SimonDanisch:
Thanks :)

PR #666, opened 2021-03-22
Clean up categorical

greimel:
This PR cleans up code and does some refactoring to make the follow-up PR #667  easier to review.

- [x] rename Continous => `Continuous`
- [x] rename categoric_fun => `categorical_fun`
- [x] use consistent arguments across `categorical_fun`
- [x] add `categorical_positions` (note the plural) to avoid recomputing `categorical_labels` for old-style `Categorical`

Please let me know how I can help making the review easier for you.

Does this need extra tests?

This
```julia
scatter(["a", "b", "c"], ["d", "e", "f"], [5, 6, 7])
```
works as expected.

image

PR #667, opened 2021-03-22
Improve Categorical trait

greimel:
Builds on #666. Introduce an improved categorical trait `HasRefPool`. If a categorical vector implements the DataAPI interface for categorical arrays then we can use `DataAPI.refpool` and `DataAPI.invrefpool` to infer the mapping from integers to labels.

With this PR one can easily change the ordering of categorical arrays.

```julia
using CairoMakie, CategoricalArrays

cat0 = rand(["lo", "mid", "hi"], 100)
cat1 = categorical(cat0, levels = ["lo", "mid", "hi"])
	

x = rand(100)
	
x .= x .- (cat0 .== "lo")
x .= x .+ (cat0 .== "hi")

categorical_ticks(cat) = AbstractPlotting.categorical_range(cat), 
		AbstractPlotting.categorical_labels(cat)

let cat = cat1
	fig = Figure()
	ax = Axis(fig[1,1])
	
	scatter!(ax, cat, x)
		
	ax.xticks = categorical_ticks(cat)
	
	fig
end
```

before:

image

after:
image


greimel:
Once #666 is merged I'll rebase this, and this will have a very clean diff.

piever:
I also think it's kind of dangerous to allow "wrong plots" as in https://github.com/JuliaPlots/Makie.jl/issues/453.

> So my idea is that categorical conversion either needs to be completely restricted to explicitly categorical arrays, which already carry the conversion to numbers within them (then the user is on their own to make sure that categorical mappings to numbers match across all plots). Or conversion needs to be "synchronized" for all categorical plots taking part in the same axis. This would mean that once a categorical plot is plotted into an axis, that axis "knows" of one mapping of categories to numbers, and can update this accordingly if new plots are added.

I'm also unsure on this one. On the one hand, it may be a bit harsh to error on `barplot(["a", "b", "c"], rand(3))` (`["a", "b", "c"]` does not even have repeated entries, so there's no real advantage to using a `CategoricalArray` there).  On the other hand, having `convert_arguments` depend on what has been plotted already on the axis seems like a huge complication. Maybe we should figure out if it would be useful for something else. I imagine the same mechanism could be recycled to implement dodging and stacking cleanly (where the exact `x` position is updated as new things are added)?

greimel:
I agree that it would be nice to handle this more cleanly. 

The problem is that right now it is *very* inconvenient to use because the order of levels is determined by `unique`, which in turn depends on the order of first occurrence.

I'd much rather have an intermediate imperfect (but strictly improving!) solution now, than wait for a few months until we've come up with the perfect solution.

(I currently need this patch for a project.)

PR #668, opened 2021-03-22
better limits

jkrumbiegel:


ffreyer:
I wonder if it would make sense to have a `ylims!(ax, lower)`. Could be nice for `hist!` for example, to align bars at the bottom.

jkrumbiegel:
Currently you would use `tightlimits!(ax, Bottom())` for that, which sets the lower automargin to 0 and then calls `autolimits!`

SimonDanisch (approved):

PR #677, opened 2021-03-28
Data inspector

ffreyer:
Very early version, but should currently work with 2D and 3D scatter, meshscatter and lines.

https://github.com/JuliaPlots/GLMakie.jl/pull/173 should get merged before this so we can easily add some fuzzyness/range to the selection. We probably also want an attribute like `pickable::Bool` so that lines etc from an `Axis` (for example) can be excluded. 

To try it out:
```julia
fig, ax, p = lines(rand(Point3f0, 10), linewidth=10)
x = AbstractPlotting.data_inspector(ax.scene)
fig
```

### DIFF
--- a/src/AbstractPlotting.jl
+++ b/src/AbstractPlotting.jl
@@ -142,10 +142,12 @@ include("stats/violin.jl")
 # Interactiveness
 include("interaction/events.jl")
 include("interaction/interactive_api.jl")
+include("interaction/inspector.jl")
 
 # documentation and help functions
 include("documentation/documentation.jl")
 include("display.jl")
+
 
 # help functions and supporting functions
 export help, help_attributes, help_arguments
@@ -207,6 +209,7 @@ export dropped_files
 export hasfocus
 export entered_window
 export disconnect!, must_update, force_update!, update!, update_limits!
+export data_inspector
 
 # Raymarching algorithms
 export RaymarchAlgorithm, IsoValue, Absorption, MaximumIntensityProjection, AbsorptionRGBA, IndexedAbsorptionRGBA
--- a/src/interaction/inspector.jl
+++ b/src/interaction/inspector.jl
@@ -46,11 +46,17 @@ end
 
 function draw_data_inspector!(inspector)
     a = inspector.attributes
-    p = text!(
+    p1 = text!(
         inspector.parent, a.display_text, 
-        position = a.position, visible = a.visible, halign = a.halign
+        position = a.position, visible = a.visible, halign = a.halign,
+        overdraw=true
     )
-    push!(inspector.plots, p)
+    p2 = scatter!(
+        inspector.parent, map(x -> [x], a.position), 
+        color = (:yellow, 0.5), strokecolor = :red, overdraw=true,
+        visible = a.visible
+    )
+    push!(inspector.plots, p1, p2)
     nothing
 end
 
--- a/src/interaction/inspector.jl
+++ b/src/interaction/inspector.jl
@@ -1,0 +1,146 @@
+struct DataInspector
+    parent::Scene
+    attributes::Attributes
+    plots::Vector{ScenePlot}
+
+    whitelist::Vector{ScenePlot}
+    blacklist::Vector{ScenePlot}
+end
+
+
+function data_inspector(
+        scene; 
+        whitelist = ScenePlot[], blacklist = ScenePlot[], range = 10
+    )
+
+    inspector = DataInspector(
+        scene,
+        Attributes(
+            display_text = " ",
+            position = Point3f0(0),
+            visible = false,
+            halign = :left
+        ),
+        ScenePlot[], whitelist, blacklist
+    )
+
+    on(events(scene).mouseposition) do mp
+        # This is super cheap
+        is_mouseinside(scene) || return false
+
+        # range requires GLMakie#173
+        plt, idx = pick(scene, mp) # , range)
+        @info idx, typeof(plt)
+        if plt === nothing
+            inspector.attributes.visible[] = false
+        else
+            show_data(inspector, plt, idx)
+        end
+    end
+
+    draw_data_inspector!(inspector)
+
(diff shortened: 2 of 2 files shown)


ffreyer:
Related:
https://github.com/JuliaPlots/Makie.jl/issues/839
https://discourse.julialang.org/t/interactive-plot-data-inspection-with-makie-a-la-plotlyjs/58095/8

ffreyer:
Just a reminder to myself - meshes are currently drawn under 2d scatter markers because those use `fxaa = false` and are therefore rendered later. I don't think that means depth tests shouldn't work though. Maybe I can fix that...

jkrumbiegel:
Would be cool if this also worked with `Axis` in that one can enable and disable the interaction, now it looks to me as if it's an "add once" kind of thing

ffreyer:
Also related: https://github.com/JuliaPlots/Makie.jl/issues/810

ffreyer:
> Would be cool if this also worked with Axis in that one can enable and disable the interaction, now it looks to me as if it's an "add once" kind of thing

I added an `enable!/disable!` now. I would like this to work with any scene so I don't want to rely on Axis interactions. 

I'm also still thinking about how to best ignore plots. With how it is right now only plots that exist when `DataInspector(fig/ax)` is called are added to the blacklist/whitelist. So you may end up being able to inspect things you generally want to ignore, or not being able inspect thing you do want to inspect. 

My favorite solution for this is still adding an attribute `pickable/inspectable`. It would allow us to set, for example, Axis plots as not inspectable by default regardless of when they are added. I think `plot!(..., inspectable = false)` is also nicer for a user than messing with a blacklist or whitelist. It would just take a lot of tiny edits to do this now.

src/basic_recipes/spy.jl
```diff
@@ -1,4 +1,4 @@
-
+# This is broken btw :)
```
ffreyer, line 1: ^

ffreyer:
I wrote some stuff to select and mark polygons. If we adjusted `contourf` to have levels in colors we could show that information.

Screenshot from 2021-04-22 04-27-52

ffreyer:
Surface now works

Screenshot from 2021-04-22 17-49-13

Maybe it would be good to finish this up now and deal with other specialization later. The only primitives that do not have a visualization are `text` (not interesting imo) and `volume` (I don't know what to show) at this point. Every other plot will fall back onto those if it doesn't have a specialization here.

src/basic_recipes/contourf.jl
```diff
@@ -141,12 +135,14 @@ function AbstractPlotting.plot!(c::Contourf{<:Tuple{<:AbstractVector{<:Real}, <:
 
     poly!(c,
         polys,
-        # colormap = c._computed_colormap,
-        # colorrange = colorrange,
+        colormap = c._computed_colormap,
+        colorrange = colorrange,
```
ffreyer, line 139: I want to highlight this because I've done more than just add `inspectable` here. I don't really use contours so maybe someone else can check that these changes are fine.

ffreyer:
I'm basically just waiting on https://github.com/JuliaPlots/GLMakie.jl/pull/177 and #696 now. Once those are merged I need to clean up some prints and that should be it.

ffreyer:
I think I also still need to make a bunch of layoutables un-inspectable.

ffreyer:
If think this should be in a merga-able state now.

Some things we should do later:
* precompilation to reduce lag
* find some way to stop 3d scenes from resizing when bounding boxes are shown

ffreyer:
Nvm, the `inspectable` attribute seems to clash with CairoMakie!?

ffreyer:
```julia
 Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] rgbatuple(c::Char)
    @ CairoMakie ~/.julia/packages/CairoMakie/GFZiw/src/utils.jl:87
  [3] draw_poly(scene::Scene, screen::CairoMakie.CairoScreen{Cairo.CairoSurfaceIOStream{UInt32}}, poly::Poly{Tuple{Vector{Vector{Point{2, Float32}}}}}, points::Vector{Point{2, Float32}}, color::Char, model::StaticArrays.SMatrix{4, 4, Float32, 16}, strokecolor::Symbol, strokewidth::Float64)
    @ CairoMakie ~/.julia/packages/CairoMakie/GFZiw/src/overrides.jl:45
  [4] (::CairoMakie.var"#58#59"{Scene, CairoMakie.CairoScreen{Cairo.CairoSurfaceIOStream{UInt32}}, Poly{Tuple{Vector{Vector{Point{2, Float32}}}}}})(points::Vector{Point{2, Float32}}, color::Char, strokecolor::Symbol, strokewidth::Float64)
    @ CairoMakie ~/.julia/packages/CairoMakie/GFZiw/src/overrides.jl:56
  [5] broadcast_foreach(::Function, ::Vector{Vector{Point{2, Float32}}}, ::Vararg{Any, N} where N)
    @ AbstractPlotting ~/work/AbstractPlotting.jl/AbstractPlotting.jl/src/utilities/utilities.jl:162
  [6] draw_poly(scene::Scene, screen::CairoMakie.CairoScreen{Cairo.CairoSurfaceIOStream{UInt32}}, poly::Poly{Tuple{Vector{Vector{Point{2, Float32}}}}}, points_list::Vector{Vector{Point{2, Float32}}})
    @ CairoMakie ~/.julia/packages/CairoMakie/GFZiw/src/overrides.jl:53
```

Could this be related to https://github.com/JuliaPlots/CairoMakie.jl/pull/159 @jkrumbiegel ?

ffreyer:
I checked a couple of plots, everything seems to be working. I also split up the docs for interactivity a bit, because I thought it was getting a bit large and fragmented for one page.

SimonDanisch (dismissed):
Thanks, this looks great :)

SimonDanisch (approved):

PR #683, opened 2021-03-31
conversion changes

SimonDanisch:
@greimel @jkrumbiegel these are the changes from #666, #667 and #619 combined quick and dirty...
Tests pass - do we actually have enough tests for the changes this covers?


greimel:
Thanks, I wasn't aware of #619. I see now where your hesitance concerning my PRs comes from @jkrumbiegel 😅

jkrumbiegel:
I think I still wanted to tweak what happens if conversion errors. Because of the loopiness of the whole thing, it could be that currently the signature after single-argument-conversion would be reported, which would be confusing for the user.

SimonDanisch:
We decided to only merge https://github.com/JuliaPlots/AbstractPlotting.jl/pull/619

PR #692, opened 2021-04-13
adjust parameters for mp4 recordings

ffreyer:
This adjusts the ffmpeg parameters for mp4s in `save()` to use `profile = high422` by default, which should fix some compatibility problems. I also made some adjustments that `profile` and `pixel_format` can be set from `save()` and `record()`.

This fixes some compatibility problems I have with videos created on linux (which defaults to "high444" encoding for me). I also have issues with mkv and webm, but I don't know how to fix those. (I assume it's also an issue of using a to recent profile/encoding!?)

### DIFF
--- a/src/display.jl
+++ b/src/display.jl
@@ -337,9 +337,7 @@ Returns a stream and a buffer that you can use, which don't allocate for new fra
 Use [`recordframe!(stream)`](@ref) to add new video frames to the stream, and
 [`save(path, stream)`](@ref) to save the video.
 """
-function VideoStream(
-        scene::Scene; framerate::Integer = 24
-    )
+function VideoStream(scene::Scene; framerate::Integer = 24)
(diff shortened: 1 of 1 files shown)


src/display.jl
```diff
@@ -519,15 +529,7 @@ extension.  Allowable extensions are:
...
 The `compression` argument controls the compression ratio; `51` is the
-highest compression, and `0` is the lowest (lossless).
-
-When `sleep` is set to `true` (the default), AbstractPlotting will
-display the animation in real-time by sleeping in between frames.
-Thus, a 24-frame, 24-fps recording would take one second to record.
-
-When it is set to `false`, frames are rendered as fast as the backend
-can render them.  Thus, a 24-frame, 24-fps recording would usually
-take much less than one second in GLMakie.
```
ffreyer, line 530: I removed this bit because I don't see `sleep` being used/passed anywhere. As far as I can tell it just goes as fast as possible by default.

SimonDanisch: Ah, yeah I forgot to clean up the docs when removing that argument

PR #696, opened 2021-04-17
Update pick functions

ffreyer:
This pr ...
* formally adds  `pick(scene, rect)`, which was already in GLMakie, but not connected and I believe not working.
* adds `pick_closest(scene, screen, xy, range)` as fallback/implementation for `pick(scene, xy, range)`.
* adds `pick_sorted(scene, screen, xy, range)` to get all plots and indices in a given range of xy, sorted by distance. Required for #677.

Requires https://github.com/JuliaPlots/GLMakie.jl/pull/177.

I checked the following with a small scatter plot
* `pick(scene, pos)`
* `pick(scene, pos, range)` with and without overloads from GLMakie
* `pick(scene, rect)`
* `pick_sorted(scene, pos, range)` with and without overloads from GLMakie

This removes what `pick(screen, rect)` for `unique(pick(scene, rect))` so it's technically a breaking change. But I think that function was broken anyway and didn't follow the usual syntax.

### DIFF
--- a/src/interaction/interactive_api.jl
+++ b/src/interaction/interactive_api.jl
@@ -137,7 +137,7 @@ sorted by distance to `xy`.
 """
 function pick_sorted(scene::SceneLike, xy, range)
     screen = getscreen(scene)
-    screen === nothing && return (nothing, 0)
+    screen === nothing && return Tuple{AbstractPlot, Int}[]
     pick_sorted(scene, screen, xy, range)
 end
 
--- a/src/interaction/interactive_api.jl
+++ b/src/interaction/interactive_api.jl
@@ -84,19 +84,104 @@ end
 
 
 """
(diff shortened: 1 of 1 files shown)


ffreyer:
With this backends need to implement:
* `pick(scene, screen, xy) -> (plot, idx)` (like before)
* `pick(scene, screen, rect) -> Matrix{Tuple{AbstractPlot, Int}}` (sort of new)

which is somewhat simpler than before.

They can also implement
* `pick_closest(scene, screen, xy, range) -> (plot, idx)`
* `pick_sorted(scene, screen, xy, range) -> distance sorted Vector{Tuple{AbstractPlot, Int}}`

if those can be done more efficiently

SimonDanisch (approved):

PR #698, opened 2021-04-19
Indent contour(3d) signatures in docstrings

mortenpi:
Currently they are rendered as text: http://makie.juliaplots.org/dev/plotting_functions/contour.html#AbstractPlotting.contour


SimonDanisch:
Thanks!

PR #702, opened 2021-04-24
small tweaks to barplot dodge formulas

piever:
This addresses the concern in https://github.com/JuliaPlots/AbstractPlotting.jl/pull/580/files#r619388750

Basically now `x_gap` is used only if the user does not pass an explicit `width`, otherwise the `width` is respected (I've added an example for this). This also allows to simplify the formulas for the "dodge" computation.

### DIFF
--- a/src/basic_recipes/barplot.jl
+++ b/src/basic_recipes/barplot.jl
@@ -13,6 +13,7 @@ $(ATTRIBUTES)
         colormap = theme(scene, :colormap),
         colorrange = automatic,
         dodge = automatic,
+        n_dodge = automatic,
         x_gap = 0.2,
         dodge_gap = 0.03,
         marker = Rect,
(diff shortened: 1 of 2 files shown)


src/basic_recipes/barplot.jl
```diff
@@ -59,12 +59,12 @@ function AbstractPlotting.plot!(p::BarPlot)
             x_unique = unique(filter(isfinite, x))
             
             if length(x_unique) == 1
-                width = 1.0
+                width = 1 - 2x_gap
```
piever, line 62: I'm respecting the previous implementation here, but maybe it could be easier to have `x_gap` default to `0.2` (rather than `0.1`) and just do `width = 1 - x_gap`. Somehow it feels natural for `x_gap` to measure the gap between bars rather than the "half gap".

piever:
Adding link to the new doc page for simplicity:

http://makie.juliaplots.org/previews/PR702/plotting_functions/barplot.html

greimel:
I didn't intend to change the default widths (for non-dodged bar plots) with my previous PR.

I am in favor of making `x_gap = 0.2` the default for consistency.

piever:
Ok, so I've updated `x_gap` to be consistent with the previous behavior (`x_gap` is now was the implicit `0.2` that we subtracted from `1` to get some spacing between bars). I've also changed the default bar width selection to use the minimum (not the mean) of the differences between `x` values, to avoid overlapping bars in the non-homogenous case.

If docs build correctly, it's good to go on my side. I can make a separate PR later to port these changes to `boxplot`, `violin` and `crossbar`.

piever:
Final tweak needed to make this usable from AoG and in complex (multi-axis) plots: I've made `n_dodge` configurable, in case some axes end up having fewer classes than others.

If there are no concerns, I was thinking of merging this in a couple of days and proceed to make a similar PR to support  dodging in `boxplot`, `crossbar`, and `violin`.

SimonDanisch:
Thanks! I guess this is ready to merge?  :)

piever:
Yes, feel free to merge!

PR #704, opened 2021-04-26
add abline

SimonDanisch:


src/makielayout/helpers.jl
```diff
@@ -448,5 +448,19 @@ Create vertical lines across `ax` at `xs` in data coordinates and `ymin` to `yma
...
 they are broadcast to calculate the final line segments.
 """
-vlines!(ax::Axis, xs; ymin = 0.0, ymax = 1.0, attrs...) = 
-    hvlines!(ax, 2, xs, ymin, ymax; attrs...)
\ No newline at end of file
+vlines!(ax::Axis, xs; ymin = 0.0, ymax = 1.0, attrs...) =
+    hvlines!(ax, 2, xs, ymin, ymax; attrs...)
+
+"""
+    abline!(axis::Axis, a::Number, b::Number; line_kw_args...)
```
piever, line 455: Mine is probably not a very insightful comment, but `abline!` is a bit of an odd name to me (I think it comes from R plotting originally). IMO `a, b` should probably be names of keyword arguments rather than in the function name (ie `line!(ax, a=0.1, b=0.2)`), and even there they are slightly unfortunate as it is not super clear which is which (I would have guessed `a` is slope and `b` is intercept).

Maybe `line!(axis; slope, intercept)` would be a clearer signature?

SimonDanisch: Well, since we can't and maybe shouldn't use keyword arguments for dispatch or changing the functionality of a function, I don't think `line!` can be used here...
The name isn't super clear I suppose, but the docs should make up for it?

SimonDanisch: Oh I guess `lines!` is the taken one and `line!` would be free... Hm ^^ :D

piever: Yes, maybe just `line!(axis, a, b)` is a good compromise. The docstring explains which is which and the function itself is easy to find by tab-completion.

SimonDanisch: Hm, I do feel like `line!` does promise quite a bit more then just a simple `x*a+b` line, and is too close for my taste to `lines!`.
Lets stay with `abline` since other packages also seem to use that name for this...

PR #705, opened 2021-04-26
Improve basic tutorial consistency and word usage

musm:
Improve consistency and word usage.

- Instead of `LinRange` use `range` consistency (more canonical)
- Consistently use `fig` for `Figure` objects
- Explain what `..` notation denotes (as this is non-standard notation not in `Base` Julia
- Improve wording in several paragraphs.


docs/src/basic-tutorial.md
```diff
@@ -20,9 +20,9 @@ For more information, have a look at [Backends & Output](@ref).
 
 Ok, now that this is out of the way, let's get started!
 
-## Importing CairoMakie
+## First plot
 
-First, we import CairoMakie, which might take a little bit of time because there is a lot to precompile. Just sit tight!
+First, we import CairoMakie.
```
ericphanson, line 25: I like the “sit tight”, it feels friendly

docs/src/basic-tutorial.md
```diff
@@ -302,19 +301,19 @@ You can switch between those workflows however you please.
...
+fig
 ```
 
-## Passing attributes to implicit Figure and Axis
+## Passing attributes to Figure and Axis
 
-For one-off plots, it's convenient to set a few axis or figure settings directly with the plotting command.
-You can do this only with the plotting functions without `!` like `lines` or `scatter`, because those always create a new axis, and can create a new figure if they are not plotting into an existing one. This is explained further under [Plot Method Signatures](@ref).
+For one-off plots, it can be convenient to set axis or figure settings directly with the plotting command.
+You can do this using the plotting functions without the `!` suffix, like `lines` or `scatter`, because thoese always create a new axis and also create a new figure if they are not plotting onto an existing one. This is explained further under [Plot Method Signatures](@ref).
```
ericphanson, line 314:
```suggestion
You can do this using the plotting functions without the `!` suffix, like `lines` or `scatter`, because these always create a new axis and also create a new figure if they are not plotting onto an existing one. This is explained further under [Plot Method Signatures](@ref).
```

docs/src/basic-tutorial.md
```diff
@@ -325,16 +324,16 @@ heatmap(randn(20, 20),
 )
 ```
 
-If you set only one attribute, be careful to do `axis = (key = value,)` (note the trailing comma), otherwise you're not making a NamedTuple but a local variable `key`.
+If you set only one attribute, be careful to do `axis = (key = value,)` (note the trailing comma), otherwise you're not creating creating a `NamedTuple` but a local variable `key`.
```
ericphanson, line 327:
```suggestion
If you set only one attribute, be careful to do `axis = (key = value,)` (note the trailing comma), otherwise you're not creating a `NamedTuple` but a local variable `key`.
```

docs/src/basic-tutorial.md
musm: The reason I removed this is that on 1.6, packages are precompiled by default, so this sentence doesn't really when loading the package.

docs/src/basic-tutorial.md
musm: Thanks!

musm:
Thanks for the review @ericphanson

ericphanson:
No problem! The changes look good to me btw, I think `range` will be more familiar to many users and `fig` is a bit more clear and explicit

docs/src/basic-tutorial.md
ericphanson: Ah true. What about

> First, we add and import CairoMakie, which might take a little bit of time because there is a lot to precompile. Just sit tight!

where the adding bit isn’t shown in code but presumably the user is often going to be adding it in order to use it. Also if we get the “auto suggest to add when you use” feature then importing could be triggering an add.

SimonDanisch:
Sounds all like good changes!
I'll leave the rest of the review to @jkrumbiegel who worte most of this ;)

musm:
Failed to push https://github.com/JuliaPlots/AbstractPlotting.jl/pull/705/checks?check_run_id=2457844443#step:7:154 ? 
Otherwise tests are passing.

SimonDanisch:
Yeah, you need to have admin rights to push the per PR preview of the docs...

PR #708, opened 2021-04-27
fix zoom for scroll values larger than +-1

ffreyer:
The zoom formula we currently have assume scroll to be -1, 0 or +1. If this is not the case the zoom factor can becomes negative which messes things up (e.g. flipping axes). This pr should fix that by using scroll steps as an exponent. 

This changes the zoom behavior slightly. For `cam2d!` the zoom in and the zoom out were different factors, meaning you didn't get back to the same point. Now you should. MakieLayout / Axis should act the same I think. `cam3d!` now (did it used to?) also returns to the same position. `cam3d_cad!` should act the same as before with scrolls of -1, 0, 1 without going crazy with larger scroll values.

### DIFF
--- a/src/camera/camera2d.jl
+++ b/src/camera/camera2d.jl
@@ -138,7 +138,7 @@ function add_zoom!(scene::SceneLike, cam::Camera2D)
         zoom = Float32(x[2])
         if zoom != 0 && ispressed(scene, zoombutton) && is_mouseinside(scene)
             pa = pixelarea(scene)[]
-            z = 1f0 + (zoom * zoomspeed)
+            z = (1f0 - zoomspeed)^zoom
             mp = Vec2f0(e.mouseposition[]) - minimum(pa)
             mp = (mp .* wscale(pa, area)) + minimum(area)
             p1, p2 = minimum(area), maximum(area)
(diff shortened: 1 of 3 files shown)


src/camera/camera2d.jl
```diff
@@ -138,7 +138,7 @@ function add_zoom!(scene::SceneLike, cam::Camera2D)
         zoom = Float32(x[2])
         if zoom != 0 && ispressed(scene, zoombutton) && is_mouseinside(scene)
             pa = pixelarea(scene)[]
-            z = 1f0 + (zoom * zoomspeed)
+            z = (1f0 - zoomspeed)^zoom
```
ffreyer, line 141: Ah, I also switched the direction of zooming here by changing + to -.  I did that because MakieLayout/Axis uses this direction for zooming. 
It can easily be changed by changing the sign of zoomspeed. Maybe we should note that somewhere

musm: This makes more sense to me than what's currently used for the zoom direction.

musm:
Tested that this fixes my problem

musm:
Who'd be good to review this?

SimonDanisch:
You tried this out right? Does it introduce any regressions?

musm:
Yes I tried it out and it fixes the issue. I did not observe any other regressions in my testing.

SimonDanisch:
Alright, thanks

PR #712, opened 2021-04-27
Refactor dodge and improve boxplot, violin and crossbar

### LINKED ISSUE #701: boxplot
Can you add an example of a boxplot in the documentation or just here for now? I can't draw boxplots in Makie and it's likely that I don't understand its syntax.

Thanks,

piever:
This PR refactors the `dodge` machinery so that it can be used without copying code for `boxplot, crossbar, violin`. It also adds example of those to the docs.

One remark: `violin` here performs horribly (at least with CairoMakie), which I think is due to the fact that KernelDensity by default returns 2048 interpolation points, so we are drawing a polygon with 2048 sides for each x value. Can I just reduce the number of interpolation points to 256? I would do that both for `violin` and `density`, as they suffer from the same issue.

Fixes #701

### DIFF
--- a/docs/src/plotting_functions/boxplot.md
+++ b/docs/src/plotting_functions/boxplot.md
@@ -25,7 +25,6 @@ AbstractPlotting.inline!(true) # hide
 xs = rand(["a", "b", "c"], 1000)
 ys = randn(1000)
 dodge = rand(1:2, 1000)
-color = ["orange", "teal", "orange", "teal", "orange", "teal"]
 
-boxplot(xs, ys, dodge = dodge, color = color, show_notch = true)
+boxplot(xs, ys, dodge = dodge, show_notch = true)
 ```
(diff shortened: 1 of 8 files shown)


piever:
(Marking as draft as the `boxplot` code can probably be cleaned up a bit here.)

piever:
Links to new docs:

- http://makie.juliaplots.org/previews/PR712/plotting_functions/boxplot.html
- http://makie.juliaplots.org/previews/PR712/plotting_functions/crossbar.html
- http://makie.juliaplots.org/previews/PR712/plotting_functions/violin.html

musm:
I definitely agree that carrying out the KS density at such a high number of grid points is very unnecessary. For example MATLAB only uses 100 points by default. I think 256 (even 128) is a reasonable number.

piever:
I've set it to `200` because I've noticed that is consistently what AbstractPlotting uses in these cases (like in `density`), even if it is not a power of `2`. In any case, we should change them all in a separate PR if we go for a different default.

PR #716, opened 2021-04-27
Add RegisterAction to help tagging

musm:
This should help automate tagging even further

image

This will update `Project.toml` to bump up the patch and trigger the Registeration bot!


SimonDanisch:
Ha, I was just wondering if something like this exists, literally a second ago while going through my github notifications :D

PR #717, opened 2021-04-28
add default plottype to three vectors

piever:
For consistency, given that `plot(rand(10), rand(10))` defaults to a scatter plot, then IMO `plot(rand(10), rand(10), rand(10))`  should default to a 3D scatter plot.


SimonDanisch:
Seems logical!

PR #719, opened 2021-04-28
formatting

piever:
A very minor thing, but I found the way the `RGBA` was specified a bit confusing to read.


SimonDanisch:
Fair enough :D Probably needs a new major version when tagged :P

PR #720, opened 2021-04-29
jk/arrows scale fix

jkrumbiegel:
- correct 2d marker rotation
- better arrows example


jkrumbiegel:
yes I saw yours, although that one works in data space right? so more suited for the 3D arrows

ffreyer:
Just in case you didn't notice - the arrow markers are still not correctly rotated with this. It becomes more visible if you have extreme aspect ratios. I don't think we'll get around changing things in GLMakie to fix this.

Maybe it would be good to implement three projection modes for all plots:
* full data space - i.e. use the projection, view etc matrices like most plots are doing
* partial pixel space - use `project(scene, ...)` on all inputs but not attributes, then assume everything in pixel space
* full pixel space - assume everything to be in pixel units from the start

This would be a case of partial pixel space, same as text. Full pixel space might be useful if you want to have some overlay without adding a new scene and messing with `clear=false`.

jkrumbiegel:
i didn't fix anything for glmakie, which needs its marker rotation with pixelspace logic straightened out. It worked correctly for cairomakie last time I tried

ffreyer:
Ah, you're right, it looks correct in CairoMakie. It's also better in GLMakie, just not perfect. I guess the bigger issue atm is the difference in markersize.

> I'm not sure about the three modes, for example strokewidth is always in pixel space, but we don't treat it in a special way. For me markers (unless they are real meshes in 3D) are the same thing, just marks that are put onto projected data positions in a size that is meaningful on the screen, but not relative to the axes.

I'm thinking about what we could easily add in GLMakie. Swapping between the projected matrices and pixelspace matrices would be easy. And it might be useful for more things than just text and scatter. For example you could handle a pixel-sized background rectangle/mesh in a scene with 3D projection by adjusting some attribute. Having the step between them is something we could do in AbstractPlotting for convenience.
If there are attributes that are in pixel space without swapping to pixel space matrices in GLMakie they would be in pixel space either way.

SimonDanisch:
I've been thinking to add support to just switch out the projection matrices more easily... I think it'd be a good idea, to have these different spaces be usable consistently

ffreyer:
I set the default 2d markersize to `20`. Not sure if that's a good default
Screenshot from 2021-05-17 16-48-31

jkrumbiegel:
maybe just leave it at `theme(scene, :markersize)`

ffreyer:
Now looks like this:
Screenshot from 2021-05-17 17-00-59

I guess we can merge once tests pass?

PR #721, opened 2021-04-29
fix default resolution and tweak sizes

jkrumbiegel:
Sets the default resolution to 800x600

Reasons:
- 4x3 is a better aspect ratio for single plots than 16:9 (which is what the defaults are most important for, one-off plots)
- 800x600 is close enough to the previous kind-of-default for linux, 960x540, so that it won't suddenly completely mess up every existing plot.
- all examples should look the same on all systems by default, the `reasonable_resolution` messes with that
- the reasonable_resolution wasn't working for linux anyway

I also adjusted font sizes, paddings etc to reach a reasonable default look. So far, everything was a bit too big with a little too much whitespace. Now, 16(px) is the default font size, which in web standards is 12pt, which is a good default font size. We should set a default `pt_per_unit` of 0.75 in CairoMakie to account for the apparent size difference. This should make pngs and svgs the same apparent size when using CairoMakie in Pluto or comparable frontends.

Before:

test_prev

After (the smaller size is hard to see here because it's just scaled):

test_now


piever:
Looks very nice! The new space hierarchy is a big readability improvement (at least for me). Now it's much clearer in plot grids that the x label belongs to the upper plot and the title to the lower plot, whereas before it was a bit confused.

jkrumbiegel:
good point, I have made the layout gaps a little bit wider by default now

grafik

PR #722, opened 2021-04-30
Only deploydocs if the DOCUMENTER_KEY env variable is non-empty

musm:
This works around the `deploydocs` failure with PRs from users without write/admin permissions, where regardless of the docs building successfully the doc CI check automatically fails.


musm:
This fixes the X (failing) doc builds from PRs created from users without admin rights.

E.g.  in https://github.com/JuliaPlots/AbstractPlotting.jl/pull/705

SimonDanisch:
Seems like a good idea, thanks! :)

PR #724, opened 2021-05-02
make docstrng of `crossbar` correct

Datseris:


Datseris:
https://makie.juliaplots.org/stable/plotting_functions/crossbar.html  the call signature should be in code

piever (approved):

PR #725, opened 2021-05-02
do not export `step!`

Datseris:
This is more like a discussion rather than a fix. In Slack I've raised the point that several Julia packages export step. The ones I know are DifferentialEquations.jl, DynamicalSystems.jl and Agents.jl. 

I'm not sure if the exported `step!` here is used somewhere, but I don't know either.

### DIFF
--- a/src/AbstractPlotting.jl
+++ b/src/AbstractPlotting.jl
@@ -230,7 +230,7 @@ export PlotSpec
 export plot!, plot
 
 
-export Stepper, step!, replay_events, record_events, RecordEvents, record, VideoStream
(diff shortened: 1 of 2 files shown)


ffreyer:
It's also not documented beyond generated docs (I thought it was mentioned elsewhere, but I don't see it) so there are probably not a lot of people that know about it/use it?

We should probably update the `Stepper` dosctring to refer to `AbstractPlotting.step!` if we make this change though. (It's in src/display.jl)

Datseris:
ok!

PR #727, opened 2021-05-03
apply different conversion pipeline to heatmap

piever:
This follows the suggestion at https://github.com/JuliaPlots/Makie.jl/issues/748, that the conversion pipeline should not be the same for `Heatmap` and `Contour, Conturf, Surface`. I've created two different traits, `DiscreteSurface` for the `Heatmap` and `ContinuousSurface` for the others (happy to switch to `CellSurfaceLike` and `VertexSurfaceLike` if those are clearer names).

For now the only difference is that `heatmap(1:3, 1:3, rand(3, 3))` gets converted to `heatmap(0.5:3.5, 0.5:3.5, rand(3, 3))`, so that now `heatmap(1:3, 1:3, rand(3, 3))` is consistent with Plots, see https://github.com/JuliaPlots/Makie.jl/issues/661.

This to me counts as "bug fix" as the current implementation of `heatmap(1:3, 1:3, rand(3, 3))`, ie split the interval `1..3` in 3 bins of length `0.66`, is IMO not correct. I'm not sure if the `(x::Vector, y::Vector, z::Matrix)` case is the only one that needs special-casing for `DiscreteSurface` or if there are others.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -249,7 +249,11 @@ function convert_arguments(P::PointBased, x::Rect3D)
     convert_arguments(P, decompose(Point3f0, x)[inds])
 end
 
-edges(v::AbstractVector) = edges(range(extrema(v)...; length=length(v)))
+function edges(v::AbstractVector)
+    l = length(v)
+    rg = l > 1 ? range(extrema(v)...; length=l) : range(extrema(v)...; step=1)
+    return edges(rg)
+end
 
 function edges(centers::AbstractRange)
     min, s, l = minimum(centers), step(centers), length(centers)
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -251,13 +251,17 @@ end
 
 function edges(v::AbstractVector)
     l = length(v)
-    rg = l > 1 ? range(extrema(v)...; length=l) : range(extrema(v)...; step=1)
-    return edges(rg)
-end
-
-function edges(centers::AbstractRange)
-    min, s, l = minimum(centers), step(centers), length(centers)
-    return range(min - s / 2, step=s, length=l + 1)
+    if l == 1
+        return [v[1] - 0.5, v[1] + 0.5]
+    else
+        # Equivalent to
+        # mids = 0.5 .* (v[1:end-1] .+ v[2:end])
+        # borders = [2v[1] - mids[1]; mids; 2v[end] - mids[end]]
+        borders = [0.5 * (v[max(1, i)] + v[min(end, i+1)]) for i in 0:length(v)]
+        borders[1] = 2borders[1] - borders[2]
+        borders[end] = 2borders[end] - borders[end-1]
+        return borders
+    end
 end
 
 function adjust_axes(::DiscreteSurface, x::AbstractVector{<:Number}, y::AbstractVector{<:Number}, z::AbstractMatrix)
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -34,8 +34,13 @@ convert_arguments(::NoConversion, args...) = args
 struct PointBased <: ConversionTrait end
 conversion_trait(x::Type{<: XYBased}) = PointBased()
 
-struct SurfaceLike <: ConversionTrait end
-conversion_trait(::Type{<: Union{Surface, Heatmap, Image}}) = SurfaceLike()
+abstract type SurfaceLike <: ConversionTrait end
+
+struct ContinuousSurface <: SurfaceLike end
+conversion_trait(::Type{<: Union{Surface, Image}}) = ContinuousSurface()
+
+struct DiscreteSurface <: SurfaceLike end
+conversion_trait(::Type{<: Heatmap}) = DiscreteSurface()
 
 function convert_arguments(T::PlotFunc, args...; kw...)
     ct = conversion_trait(T)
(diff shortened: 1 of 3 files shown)


src/conversions.jl
```diff
@@ -248,6 +249,21 @@ function convert_arguments(P::PointBased, x::Rect3D)
...
 end
 
+edges(v::AbstractVector) = edges(range(extrema(v)...; length=length(v)))
+
+function edges(centers::AbstractRange)
+    min, s, l = minimum(centers), step(centers), length(centers)
+    return range(min - s / 2, step=s, length=l + 1)
+end
+
+function adjust_axes(::DiscreteSurface, x::AbstractVector{<:Number}, y::AbstractVector{<:Number}, z::AbstractMatrix)
```
piever, line 259: This is the only scenario where the conversion behavior differs between discrete and continuous surfaces, not sure if there are others that I'm missing.

ffreyer:
For the `Vector` cases it might be good to port what I have in https://github.com/JuliaPlots/GLMakie.jl/pull/137. Specifically https://github.com/ffreyer/GLMakie.jl/blob/6e7c0972d92615b33c1dd6ed2084c18d492babf0/src/drawing_primitives.jl#L326:
```julia
# Equivalent to
# mids = 0.5 .* (v[1:end-1] .+ v[2:end])
# borders = [2v[1] - mids[1]; mids; 2v[end] - mids[end]]
borders = [0.5 * (v[max(1, i)] + v[min(end, i+1)]) for i in 0:length(v)]
borders[1] = 2borders[1] - borders[2]
borders[end] = 2borders[end] - borders[end-1]
```

This would be ignored by the current CairoMakie and GLMakie version, but I could fix up my pr pretty quickly I think.

ffreyer:
> I'm not sure if the `(x::Vector, y::Vector, z::Matrix)` case is the only one that needs special-casing for `DiscreteSurface` or if there are others.

I think so. Intervals are just boundary values and ranges always have a set step size afaik. So both generate equally sized cells. Vectors may create different sizes. Matrices could create different shapes, so they would need to transform to polygons/a mesh which we don't support.

piever:
> For the `Vector `cases it might be good to port what I have in JuliaPlots/GLMakie.jl#137. 
> ```julia
> borders = [0.5 * (v[max(1, i)] + v[min(end, i+1)]) for i in 0:length(v)]
> borders[1] = 2borders[1] - borders[2]
> borders[end] = 2borders[end] - borders[end-1]
> ```

Ah, my `edges` function does the same but assumes uniform spacing. Instead the correct thing would be to take as "inner borders" the midpoints between the initial values, and extrapolate the outer borders?

I can definitely change the code to do that, possibly also allowing the `1 x 1` case, with default width `1` (it seems silly but could in principle happen when the vectors come from categorical data).

ffreyer:
> Instead the correct thing would be to take as "inner borders" the midpoints between the initial values, and extrapolate the outer borders?

That's my idea of it at least. 

I'll see if I can update the GLMakie and CairoMakie pr's for this, but it may take me till the weekend. If the vector versions look alright we could merge this earlier I'd say.

piever:
Thanks for updating the PRs, that'd be great!

As for "mergeability", I think it's fine, because this PR converts the "`x, y` of lenght `N`, `z` of size `N x N`" case to the  "`x, y` of lenght `N + 1`, `z` of size `N x N`" case. The latter is already supported in GLMakie and CairoMakie. For example, the [heatmap from the docs](https://makie.juliaplots.org/previews/PR727/plotting_functions/heatmap.html) built fine with the ticks correctly in the middle of the bin.

ffreyer:
This used to work before but doesn't here:
```julia
fig, ax, p = heatmap(0..1, 1:10, rand(10, 10))
ERROR: There was no `AbstractPlotting.convert_arguments` overload found for
the plot type Heatmap{Tuple{IntervalSets.ClosedInterval{Int64}, UnitRange{Int64}, Matrix{Float64}}}, or its conversion trait AbstractPlotting.DiscreteSurface().
The arguments were:
(IntervalSets.ClosedInterval{Int64}, UnitRange{Int64}, Matrix{Float64})

To fix this, define `AbstractPlotting.convert_arguments(::Heatmap{Tuple{IntervalSets.ClosedInterval{Int64}, UnitRange{Int64}, Matrix{Float64}}}, ::IntervalSets.ClosedInterval{Int64}, ::UnitRange{Int64}, ::Matrix{Float64})`.
```

ffreyer:
I updated the GLMakie pr btw

piever:
> This used to work before but doesn't here:
> 
> ```julia
> fig, ax, p = heatmap(0..1, 1:10, rand(10, 10))
> ERROR: There was no `AbstractPlotting.convert_arguments` overload found for
> the plot type Heatmap{Tuple{IntervalSets.ClosedInterval{Int64}, UnitRange{Int64}, Matrix{Float64}}}, or its conversion trait AbstractPlotting.DiscreteSurface().
> The arguments were:
> (IntervalSets.ClosedInterval{Int64}, UnitRange{Int64}, Matrix{Float64})
> 
> To fix this, define `AbstractPlotting.convert_arguments(::Heatmap{Tuple{IntervalSets.ClosedInterval{Int64}, UnitRange{Int64}, Matrix{Float64}}}, ::IntervalSets.ClosedInterval{Int64}, ::UnitRange{Int64}, ::Matrix{Float64})`.
> ```
> 
> Mixing vectors and ranges works fine though

I actually get the same error on master, maybe it's better to fix it in a separate PR if we decide to support it?

ffreyer:
Oh you're right. Nvm then, let's not make this more than it needs to be.

PR #728, opened 2021-05-03
Allow vector offsets in text

ffreyer:
This doesn't break anything as far as I can tell, but it allows you to do some hacky super- and subscript as well as (one character) fractions.

```julia
scene = Scene()
text!(
    scene,
    ["kx", "π_2"],
    position = [Point3f0(1, 0.5, 0.5), Point3f0(0.5, 1, 0.25)],
    offset = [
        [Vec2f0(0), Vec2f0(0, -5)],
        [Vec2f0(0, 10), Vec2f0(-11, 11), Vec2f0(-22, -11)],
    ]
)
scene |> display
```

Screenshot from 2021-05-03 23-19-34

### DIFF
--- a/src/layouting/layouting.jl
+++ b/src/layouting/layouting.jl
@@ -231,12 +231,15 @@ function glyph_positions(str::AbstractString, font_per_char, fontscale_px, halig
         reduce(vcat, map(line -> [l.hadvance for l in line], lineinfos)))
 end
 
-function preprojected_glyph_arrays(string::String, position::VecTypes, glyphlayout::AbstractPlotting.Glyphlayout, font, textsize, space::Symbol, projview, resolution, offset::VecTypes, transfunc)
+
+function preprojected_glyph_arrays(
+        string::String, position::VecTypes, glyphlayout::AbstractPlotting.Glyphlayout, 
+        font, textsize, space::Symbol, projview, resolution, offset::VecTypes, transfunc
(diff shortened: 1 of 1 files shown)


ffreyer:
Note that explicitly supplying offsets messes with alignment. (I.e. it overwrites what is set via `align`)

jkrumbiegel:
I have a PR for this as well, how does yours break alignment? Mine doesn't, the alignment is just relative to the offset anchor. I was just looking for it but I might not have pushed it yet

ffreyer:
Huh, it doesn't, actually. I thought it shifted my anchor point when I was messing around with it before, but I guess I was mistaken.

SimonDanisch:
@jkrumbiegel your pr was merged and for single offsets, right?
https://github.com/JuliaPlots/AbstractPlotting.jl/pull/703

jkrumbiegel:
haha yeah it seems I just forgot that I merged that one, oh man

ffreyer:
I guess I should add the example as a test before this gets merged?

ffreyer:
Might be a good idea to add the same for `textsize` here. It works per character with one string, but not with an array of strings.

ffreyer:
The test now checks:

Screenshot from 2021-05-04 18-54-18

SimonDanisch (approved):

PR #732, opened 2021-05-06
fix theming of recipes

SimonDanisch:
It seems, the ability to theme recipes by setting:
`set_theme!(my_recipe= (attributes...,))` 
Has been broken at some point...

### DIFF
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -281,7 +281,7 @@ $(ATTRIBUTES)
 end
 
 function plot!(plot::Text)
-    
+
     # attach a function to any text that calculates the glyph layout and stores it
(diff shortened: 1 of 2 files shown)


piever:
Does this also allow to theme a specific attribute for all recipes?

For example, could one say `set_theme!(Theme(strokecolor=:white))` and have a white strokecolor on all recipes that support this attribute?

At the moment if one does

```julia
julia> set_theme!(strokecolor=:white)

julia> scatter(rand(10), rand(10))
```

it is not respected.

SimonDanisch:
Hm, I thought there would were some rules like that... This whole code needs a thorough clean up :D

PR #733, opened 2021-05-06
jk/gradient density

jkrumbiegel:
https://makie.juliaplots.org/previews/PR733/plotting_functions/density.html

- add mirrored even color mode for band
- switch density to band so gradient colors can work
- add gradient density examples
- add to density docstring
- use :value instead of :x


docs/src/plotting_functions/density.md
```diff
@@ -66,3 +66,46 @@ end
 
 f
 ```
+
+#### Gradients
+
+You can color density plots with gradients, either along the `:value` or `:density` dimension.
+In the horizontal default case, `:value` means along the horizontal and `:density` along the vertical.
```
SimonDanisch, line 73: Another word for x and y (thinking of https://github.com/JuliaPlots/Makie.jl/issues/926 here).
Should we just always use x/y? By always using x/y even though the axes have different names, people don't even need to look this up ;) 

If we still decide to go with this, we should name the arguments in the recipe call like this:
```julia
@recipe(Density, value, density) do scene
...
```

briochemc:
This works like a charm for me! thank you so much!

PR #734, opened 2021-05-07
Added method example to axislegend

KronosTheLate:
In this PR, I added the method example `axislegend(ax, args...; position = (1, 1), kwargs...)`

The motivation is that I went to try the function just now, and initially started inputing numbers on the size-scale of pixels, because i was not familiar with halign and valign, and the fact that they go from 0 to 1. So I figured that adding this example would potentialy clarify thing for people like me.


SimonDanisch:
That seems like a good addition!

SimonDanisch (approved):

PR #740, opened 2021-05-11
Update band.jl doc string

mschauer:


src/basic_recipes/band.jl
```diff
@@ -1,8 +1,9 @@
 """
-band(x, ylower, yupper; kwargs...)
-band(lower, upper; kwargs...)
+    band(x, ylower, yupper; kwargs...)
+    band(lower, upper; kwargs...)
 
-Plots a band from `ylower` to `yupper` along `x`.
+Plots a band from `ylower` to `yupper` along `x`. The form `band(lower, upper)` plots a [ruled surface](https://en.wikipedia.org/wiki/Ruled_surface)
```
briochemc, line 5: `lower` and `upper` may be a bit confusing in the second case. What about calling them `start_points` and `end_points` instead? Maybe this would make it clearer that this call expects `Point`s? 🤷

PR #742, opened 2021-05-11
fix buggy menu interaction

ffreyer:


ffreyer:
Fixes two problems:

* The `Menu` reacting to clicks where the menu items would be. This was fixed in the first commit by rolling back some performance tweaks, which are added again in the second commit.
* When selecting the last item in a menu and reducing the number of options, the menu breaks. This happened because the gridlayout triggered an update before the index was adjusted. It's now silently (i.e. via `observable.val`) set to 0 before that happens.

Performance tweaks:
* From PriorityObservables: Check if mouse is in the bbox calculated by the gridlayout instead of using `pick`. This means we no longer fetch a texture from the gpu.
* Removed some refs around the arrays of boxes/labels. Instead of rebuilding that array it is now mutated.

I'm checking things with this example:

```julia
using GLMakie
fig = Figure()
menu = Menu(fig, options = ["viridis", "heat", "blues"])
funcs = [sqrt, x->x^2, sin, cos]
menu2 = Menu(fig, options = zip(["Square Root", "Square", "Sine", "Cosine"], funcs))
fig[1, 1] = vgrid!(
    Label(fig, "Colormap", width = nothing),
    menu,
    Label(fig, "Function", width = nothing),
    menu2;
    tellheight = false, width = 200)
ax = Axis(fig[1, 2])
func = Node{Any}(funcs[1])
ys = @lift($func.(0:0.3:10))
scat = scatter!(ax, ys, markersize = 10px, color = ys)
cb = Colorbar(fig[1, 3], scat)
on(menu.selection) do s
    scat.colormap = s
end
on(menu2.selection) do s
    func[] = s
    autolimits!(ax)
end
menu2.is_open = true
fig |> display
```

PR #746, opened 2021-05-15
add rotation/s to Billboard

ffreyer:
Requires JuliaPlots/GLMakie.jl/pull/189, JuliaPlots/CairoMakie.jl#160 and JuliaPlots/WGLMakie.jl#99

Adds an angle to `Billboard` to allow for in-plane rotations. Should help with #720.

### DIFF
--- a/src/conversions.jl
+++ b/src/conversions.jl
@@ -752,7 +752,8 @@ function convert_attribute(c::Tuple{T, F}, k::key"color") where {T, F <: Number}
     RGBAf0(Colors.color(to_color(c[1])), c[2])
 end
 
-convert_attribute(c::Billboard, ::key"rotations") = Quaternionf0(0, 0, 0, 1)
+convert_attribute(b::Billboard{Float32}, ::key"rotations") = to_rotation(b.rotation)
+convert_attribute(b::Billboard{Vector{Float32}}, ::key"rotations") = to_rotation.(b.rotation)
 convert_attribute(r::AbstractArray, ::key"rotations") = to_rotation.(r)
 convert_attribute(r::StaticVector, ::key"rotations") = to_rotation(r)
 
--- a/src/types.jl
+++ b/src/types.jl
@@ -296,10 +296,18 @@ func2type(f::Function) = Combined{f}
 
 
 """
+    Billboard([angle::Real])
+    Billboard([angles::Vector{<: Real}])
+
 Billboard attribute to always have a primitive face the camera.
 Can be used for rotation.
 """
-struct Billboard end
+struct Billboard{T <: Union{Float32, Vector{Float32}}}
+    rotation::T
+end
+Billboard() = Billboard(0f0)
+Billboard(angle::Real) = Billboard(Float32(angle))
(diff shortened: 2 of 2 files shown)


ffreyer:
Tests passed locally for me with the GLMakie pr

ffreyer:
Using #720 with
```julia
rotations = lift(scene.camera.projectionview, scene.px_area, headstart) do pv, pxa, hs
    proj_rotations = map(hs) do (start, stop)
        pstart = project(scene, start)
        pstop = project(scene, stop)
        diff = pstop - pstart
        angle = acos(diff[2] / norm(diff))
        angle = ifelse(diff[1] > 0, 2pi - angle, angle)
    end
    Billboard(proj_rotations)
end
```
(and a pixel-scale arrowsize) results in correct rotations + no distortion
Screenshot from 2021-05-15 17-31-37

CairoMakie can't deal with it though...

ffreyer:
https://github.com/JuliaPlots/CairoMakie.jl/pull/160 should fix that

jkrumbiegel:
can we use this with a marker_offset that is applied before rotation? then we could shift the arrow tips slightly forward or backward, which could help a lot when making arrows that point to exact locations

ffreyer:
WGLMakie also needs changes for this to work. It currently ignores rotations with `Billboard{Float32}` and errors with `Billboard{Vector{Float32}}`.

> can we use this with a marker_offset that is applied before rotation? then we could shift the arrow tips slightly forward or backward, which could help a lot when making arrows that point to exact locations

Yes.
```julia
angles = collect(range(0, 2pi, length=37))
ps = [Vec2f0(cos(a), sin(a)) for a in angles]
scatter(ps, marker='▲', rotations=Billboard(angles .- pi/2), marker_offset = Vec2f0(-4, 10))
```
Screenshot from 2021-05-15 19-39-05

jkrumbiegel:
Ah cool, but what is the horizontal shift for in this case? I don't see it in the image, unless you're using it to shift the marker? That points to another problem, that it's pretty annoying having to center glyph markers manually via marker offset. In cairomakie I actually center the glyphs to their boundingbox center, because I think for scatters you obviously want the glyph centered, not positioned in the usual text way. But this behavior is not like GLMakie so we should decide something there. Same for scale btw, glyph scatters are scaled weirdly in glmakie unless correcting manually to un-square the glyph

ffreyer:
Yes, it's to center the marker. It throws `Vec2f0(-0.5markersize, -0.5markersize)` in there by default. I guess GLMakie does `rotation * translation * scale * Rect(0, 0, 1, 1)`

ffreyer:
> WGLMakie also needs changes for this to work. It currently ignores rotations with `Billboard{Float32}` and errors with `Billboard{Vector{Float32}}`.

Should be fine with JuliaPlots/WGLMakie.jl#99

jkrumbiegel:
hm okay we should assimilate the two approaches into one clear version that's documented and unit-tested. this nitty gritty stuff with rotation, translation, scaling, the order of the three, glyph offsets, etc, can make or break a detail oriented plot

SimonDanisch (approved):

PR #747, opened 2021-05-16
Switch to using `Artifacts` for `assets/`

MichaelHatherly:
Fixes https://github.com/JuliaPlots/Makie.jl/issues/949

Makes `AbstractPlotting` more amenable to inclusion in a relocated system image where the original source directories may not exist anymore.

Artifact creation is currently a manual process in this PR. Is the assets directory likely to change regularly enough that finding a way to automate the artifact creation process would be worth the extra effort, or is a manual step sufficient here?

CI _should_ fail on this pretty early since the referenced artifact URL does not exist. That'll need a maintainer to upload the built artifact first. Locally `test` seemed to work fine.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,6 @@ phases.0001_print.jpg
 **/._*
 Manifest.toml
 !docs/Manifest.toml
+/build
 
 \.DS_Store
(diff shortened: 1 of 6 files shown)


SimonDanisch:
Thanks :) 
It should only rarely need updating, so a well documented upload script should suffice!

MichaelHatherly:
The uploading script should probably be taken for a spin by someone with commit access just so we know it's actually doing the right thing.

PR #748, opened 2021-05-17
Added a title which displays time

KronosTheLate:
I wanted to add a title to show the time during an animation. But one can not @lift an observable in a string-interpolation it seems - or at least, I could not figure out how. So I added this functionality to the example, because I think it is a very normal want from an animation, and a less-than-obvious solution.

The animation file also needs updating. I don't know where the file lives, but I attach the animation produced from the new code. Hopefully someone can help me put it in the right place ^_^

https://user-images.githubusercontent.com/61620837/118468449-116ce200-b705-11eb-97b0-e97f3f735d2c.mp4

### DIFF
--- a/docs/src/animation.md
+++ b/docs/src/animation.md
@@ -78,7 +78,7 @@ xs = range(0, 7, length=40)
 ys_1 = @lift(sin.(xs .- $time))
 ys_2 = @lift(cos.(xs .- $time) .+ 3)
 
-fig  = lines(xs, ys_1, color = :blue, linewidth = 4)
+fig , ax, line = lines(xs, ys_1, color = :blue, linewidth = 4)
 scatter!(xs, ys_2, color = :red, markersize = 15)
 
 framerate = 30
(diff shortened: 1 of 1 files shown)


docs/src/animation.md
```diff
@@ -78,14 +78,15 @@ xs = range(0, 7, length=40)
 ys_1 = @lift(sin.(xs .- $time))
 ys_2 = @lift(cos.(xs .- $time) .+ 3)
 
-fig  = lines(xs, ys_1, color = :blue, linewidth = 4)
+fig , ax, line = lines(xs, ys_1, color = :blue, linewidth = 4)
```
jkrumbiegel, line 81: This is how you lift within string interpolation
```julia
fig  = lines(xs, ys_1, color = :blue, linewidth = 4,
    axis = (title = @lift("t = $(round($time, digits = 1))"),))
```

jkrumbiegel (changes requested):

jkrumbiegel:
the animation auto-regenerates when merged

SimonDanisch:
I guess the question is, should we use this to show how to interpolate strings with `@lift`, or how to update the title directly? :D

jkrumbiegel:
ok I've just added the `@lift` version myself, just because we're trying to get PRs to zero to move some stuff around :) thanks for the PR!

PR #749, opened 2021-05-17
dont use kw for lift

SimonDanisch:
This shaves off 1s from ttfp 🤷


ffreyer:
Oh, perhaps it's also worth changing this for `on(f, o::PriorityObservables; priority)`?

SimonDanisch:
That's probably not as bad, as the type isn't that important ;) Maybe we can type it though...
But I did note, that a lot of method invalidations happen because of GridLayout + Makie overloading `on` ... Maybe that's worse ;) But I'm not sure how to fix that

